PR #29, opened 2015-02-03
Allow non-HTML Opium responses to add custom HTTP headers

avsm:
Previously the Response_helper module ignored any custom HTTP
headers passed through to Json, XML or HTML responses.  We now
preserve these and just add a Content-Type header.


rgrinberg:
Thanks!

PR #31, opened 2015-02-08
Generate Content-Type headers for static file serving

avsm:
This uses the `magic-mime` library to detect the filename and
generate a sensible MIME type from the `mime.types` database.
It is particularly important to generate these for Internet Explorer
support, since IE10 seems to ignore `.css` files that do not send
an appropriate MIME header.  With this patch, IE10 rendering works
great using an Opium rendering stack.

### DIFF
--- a/_oasis
+++ b/_oasis
@@ -41,7 +41,8 @@ Library opium
                 humane_re,
                 ezjsonm,
                 base64,
-                cmdliner
+                cmdliner,
+                magic-mime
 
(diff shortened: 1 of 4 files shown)


rgrinberg:
You have IE 10?! :astonished: 

But makes sense. Thanks.

avsm:
I collared an MSR person who had a Windows phone :-)

PR #37, opened 2015-06-15
Split unix 2

rgrinberg:
@raboof what do you think?

I undid the backwards incompatible changes and tried to do them in a backwards compatible way instead. I've also renamed some stuff but I'm still unhappy with the general structure. I think I will just get rid of oasis to and write my own META file to get the kind of findlib/directory structure I want.

Also, I think we need a different underlying type for `Opium_kernel.App.t` and `Opium.App.t`. For example, I'm not sure that the port parameter makes sense for mirage based applications. I think all you need is a `Flow` to run an application.


raboof:
Changes look good to me. 

I agree the 'port' could move down to the unix/mirage-specific part.

raboof:
Are the travis failures a known issue?

PR #41, opened 2015-08-29
Add expiration parameter to cookies

venator:
This patch adds a `?expiration : Cohttp.Cookie.expiration` parameter to `Cookie.set` and `Cookie.set_cookies` functions.

This works by using a new Env for response cookies (before, the same Env than request cookies was uesd), with a cookie of type `(string * string * Cohttp.Cookie.expiration)`.

This patch doesn't break `Cookie` interface since the new parameter is optional, and default to `Session`.

### DIFF
--- a/opium/cookie.ml
+++ b/opium/cookie.ml
@@ -56,14 +56,14 @@ let set_cookies resp cookies =
 
 let set resp ~key ~data = set_cookies resp [(key, data)]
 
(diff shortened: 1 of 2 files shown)


venator:
Actually, it would probably be smarter to set cookie expiration on a per-cookie basis.

rgrinberg:
Makes sense. Thank you. I'll add some tests.

PR #42, opened 2015-10-01
Add SSL options to server

venator:
This patch adds the following options to enable SSL when starting the server:

```
-s VAL, --ssl-cert=VAL
SSL certificate file

-k VAL, --ssl-key=VAL
SSL key file
```

### DIFF
--- a/rock/opium_rock.mli
+++ b/rock/opium_rock.mli
@@ -105,5 +105,5 @@ module App : sig
 
   val create : ?middlewares:Middleware.t list -> handler:Handler.t -> t
 
-  val run : t -> port:int -> unit Lwt.t
+  val run : t -> port:int -> ssl:((string * string) option) -> unit Lwt.t
 end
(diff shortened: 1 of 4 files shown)


rock/opium_rock.mli
```diff
@@ -105,5 +105,5 @@ module App : sig
 
   val create : ?middlewares:Middleware.t list -> handler:Handler.t -> t
 
-  val run : t -> port:int -> unit Lwt.t
+  val run : t -> port:int -> ssl:((string * string) option) -> unit Lwt.t
```
rgrinberg: Can we include the polymorphic variant labels here otherwise it's a bit hard to understand what's being passed.

Also, what about making `~ssl` optional `?ssl` and putting it as the first argument to maintain backwards compat.

Other than that, looks good.

venator: I've pushed changes which should fix that.

rgrinberg:
Looks good.

Issue #40, opened 2015-08-17
core_kernel dependency?


samoht:
Just wondering if the dependency towards core_kernel is really necessary? Seems that you only need an universal map implementation.

rgrinberg:
I agree. However this has been a low priority for me until someone is actually affected by this since it's kind of a large and boring change.

I'm not aware of any standalone universal map implementations so I've also been kind of waiting for one to pop up on opam. I'm happy turn the gears on this if someone needs this however.

samoht:
That's not urgent at all, but I am starting to use it [in some of my new projects](https://github.com/samoht/ciso) (and I consider to maybe using it in Irmin in a so-no-distant future). There are few implementation or universal map in the various libraries from Daniel, but there is no stand-alone one at the moment.

hcarty:
opium without Core_kernel required and a stand-alone universal map implementation both sound like wonderful things to have.

rgrinberg:
I think @dbuenzli has some univ map implementations lying around.

Any chance we'll see them released?

rgrinberg:
I think @dbuenzli has some univ map implementations lying around.

Any chance we'll see them released?

dbuenzli:
> Any chance we'll see them released?

Yeah sure, here's a release: 

```ocaml
(** Dictionaries.

    A dictionary is a set of {{!keys}keys} mapping to typed values. *)
module Dict : sig

  (** {1:keys Keys} *)

  type 'a key
  (** The type for dictionary keys whose lookup value is ['a]. *)

  val key : unit -> 'a key
  (** [key ()] is a new dictionary key. *)

  (** {1:dict Dictionaries} *)

  type t
  (** The type for dictionaries. *)

  val empty : t
  (** [empty] is the empty dictionary. *)

  val is_empty : t -> bool
  (** [is_empty d] is [true] iff [d] is empty. *)

  val mem : 'a key -> t -> bool
  (** [mem k d] is [true] iff [k] has a mapping in [d]. *)

  val add : 'a key -> 'a -> t -> t
  (** [add k v d] is [d] with [k] mapping to [v]. *)

  val rem : 'a key -> t -> t
  (** [rem k d] is [d] with [k] unbound. *)

  val find : 'a key -> t -> 'a option
  (** [find d k] is [k]'s mapping in [d], if any. *)

  val get : 'a key -> t -> 'a
  (** [get k d] is [k]'s mapping in [d].

[...]

PR #47, opened 2016-02-21
remove core_kernel

c-cube:
This is currently *work in progress*, do not merge yet. Purpose: solve #40.

I added a `opium_kernel/opium_umap` (temporarily) to discuss the API and get one that works sufficiently well. We will fork it into its own library once we reach agreement on its interface (it is already a functor).

`Opium_misc`  now contains some utils to replace parts of `core_kernel` (missing functions, etc.). I also removed the `Info` stuff and replaced it with mere strings. I kept the sexplib stuff (a bit annoying to rewrite though). Tests do not compile yet.

### DIFF
--- a/.merlin
+++ b/.merlin
@@ -6,11 +6,12 @@ S examples
 B _build/**
 
 PKG base64
-PKG core_kernel
 PKG lwt
 PKG lwt.unix
 PKG cohttp.lwt
 PKG cmdliner
+PKG uri
+PKG sexplib
 PKG oUnit
 PKG ezjsonm
 PKG sexplib.syntax
--- a/OMakefile
+++ b/OMakefile
@@ -35,7 +35,7 @@ OCAMLFLAGS = -g -bin-annot -thread
 
 OCAMLFINDFLAGS += -syntax camlp4o
 OCAMLPACKS[] +=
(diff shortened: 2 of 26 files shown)


c-cube:
It compiles and, apparently, tests pass:

> ..............
Ran: 14 tests in: 0.00 seconds.
OK
*** omake: done (0.07 sec, 0/11 scans, 1/39 rules, 0/211 digests)

So `Opium_misc` is now a small subset of Core :-). There might be imprecisions in the translation, so it has to be reviewed ­— I don't know much about Core.

c-cube:
I will update the PR once @dbuenzli releases hmap.

rgrinberg:
I wonder if there's a need to wait for hmap, I'd really like to have this change in.

rgrinberg:
OK I just tested this and the tests pass too.

I will wait for @dbuenzli to make an hmap release before pushing out 0.15.0

c-cube:
HMap is released!

c-cube:
@rgrinberg so, how about a release? I can update to use `hmap` if you wish

rgrinberg:
@c-cube master already uses `hmap`. I will look into releasing this week.

Issue #49, opened 2016-03-29
Unable to build against latest master


hcarty:
OCaml 4.02.3 on OSX with `opam pin add opium --dev-repo`:

Trying to compile this code:

```ocaml
open Opium.Std

let () =
  App.empty
  |> get "/" (fun _req -> `String "Hello world" |> respond')
  |> App.cmd_name "HW"
  |> App.run_command
```
I get:
```
$ ocamlbuild -use-ocamlfind -package opium.unix o.native
+ ocamlfind ocamlopt -linkpkg -package opium.unix o.cmx -o o.native
File "_none_", line 1:
Error: No implementations provided for the following modules:
         Cohttp_lwt_body referenced from /Users/hcarty/.opam/pipeline/lib/opium/opium_kernel.cmxa(Opium_misc),
           /Users/hcarty/.opam/pipeline/lib/opium/opium_kernel.cmxa(Opium_rock),
           /Users/hcarty/.opam/pipeline/lib/opium/opium.cmxa(Opium_app)
         Lwt_log referenced from /Users/hcarty/.opam/pipeline/lib/opium/opium.cmxa(Opium_debug),
           /Users/hcarty/.opam/pipeline/lib/opium/opium.cmxa(Opium_app)
         Cohttp_lwt_unix referenced from /Users/hcarty/.opam/pipeline/lib/opium/opium.cmxa(Opium_app),
           /Users/hcarty/.opam/pipeline/lib/opium/opium.cmxa(Opium_static_serve)
         Lwt_main referenced from /Users/hcarty/.opam/pipeline/lib/opium/opium.cmxa(Opium_app)
Command exited with code 2.
```

rgrinberg:
Fixed in #50

PR #50, opened 2016-03-29
Update META dependencies

hcarty:
This should allow compilation against master


hcarty:
This (should) fix #49

rgrinberg:
Thanks. My next goal is auto generate these so hopefully we won't have this.

PR #53, opened 2016-05-04
update readme

c-cube:
- remove references to corebuild
- put a curl command to query the hello world example

on a side note, I didn't succeed in querying the `put` endpoint with curl. Any advice on that? Keep getting 404.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,8 +87,15 @@ let _ =
 
 compile with:
 ```
-$ corebuild -pkg opium,cow.syntax hello_world.native
+$ ocamlbuild -pkg opium.unix hello_world.native
 ```
+
+and then call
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -134,7 +141,7 @@ let _ = App.empty
 Compile with:
 
 ```
-$ corebuild -pkg opium middleware_ua.native
+$ ocamlbuild -pkg opium.unix middleware_ua.native
```
rgrinberg: `-use-ocamlfind` is necessary here I think

c-cube: I think ocamlfind support is enabled by default in a modern ocamlbuild; this does work at least on my computer.

PR #66, opened 2017-03-19
enable ssl when it's set only through api but not cli

sevenEng:
Command line options about ssl will always write over `app.ssl`. So when it has already been set by `App.ssl` function call, but not given any information through command line, the ssl will not be enabled (evaluated to `None` eventually).

In this patch, preference is given to command line option when both are not `None`.

### DIFF
--- a/opium/opium_app.ml
+++ b/opium/opium_app.ml
@@ -135,7 +135,7 @@ let print_routes_f routes =
         (data
          |> List.map ~f:Cohttp.Code.string_of_method
          |> String.concat " "))
-    routes_tbl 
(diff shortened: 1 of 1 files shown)


sevenEng:
Hi @rgrinberg !
Just check if there will be a plan of new release to include this, we are currently using a fork and a git pin to have this. With a probable new release, it would bring us much convenience when managing some repos. Thanks!

rgrinberg:
I will make a release as soon as cohttp 0.99 is  released. I'm expecting a
release of that in the upcoming days

Issue #57, opened 2016-06-03
Encoding error on get cookie?


liweijian:
Testing code

``` ocaml
open Opium.Std

let all_cookies = get "/cookies" begin fun req ->
  let cookies = req
                |> Cookie.cookies
                |> List.map ~f:(fun (k,v) -> k ^ "=" ^ v)
                |> String.concat "\n"
  in
  `String (Printf.sprintf "<pre>%s</pre>" cookies) |> respond |> return
end

let app = 
  App.empty
  |> all_cookies
  |> middleware Cookie.m

let _ =
  app |> App.run_command
```

Compile and run as

> corebuild -pkg opium.unix get_cookie.native ; ./get_cookie.native -p 9001 -d

Testing by cUrl:

``` bash
$ curl -v --cookie "name=li" http://127.0.0.1:9001/cookies       
*   Trying 127.0.0.1...
* Connected to 127.0.0.1 (127.0.0.1) port 9001 (#0)
> GET /cookies HTTP/1.1
> Host: 127.0.0.1:9001
> User-Agent: curl/7.43.0
> Accept: */*
> Cookie: name=li
>
< HTTP/1.1 200 OK
< content-length: 17
<
* Connection #0 to host 127.0.0.1 left intact
<pre>���=��</pre>~                                                                                                           ⋊> 
```

It has been changed to `��`.

rgrinberg:
@tmattio I think the new cookie module fixes this?

tmattio:
Yes, we don't decode/encode cookies in any way with the new module (except the signature of course, but that's unrelated)

PR #74, opened 2017-12-13
Do not use or assume Base 64 encoding of cookies

malthe:
Also, only cookie values should be percent-decoded. The keys should be
using the safe alphabet as per the spec.

This fixes issue #57.


rgrinberg:
Thanks!

PR #75, opened 2017-12-17
Fix README

bramford:
The `ocamlbuild` commands when executed as per the README.md result in failure:

```
$ ocamlbuild -pkg opium.unix hello_world.native
ocamlfind: Package `opium.unix' not found      
Cannot run Ocamlfind. 
```

The correct package name acquired by opam is `opium`.

I also felt it helpful to remove the shell prompt from the example commands to make the format of them copy-paste friendly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ library will always be developed against OPAM version.
 The latest stable version is available on opam
 
 ```
-$ opam install opium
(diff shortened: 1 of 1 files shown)


rgrinberg:
Thanks. I should probably replace those with jbuilder instructions.

PR #76, opened 2018-04-13
Add caching capabilities to static middleware

mattjbray:
This PR adds an optional `etag_of_fname` argument to `Middleware.static`. When provided, the middleware will add the `ETag` header to responses.

If the request includes the `If-None-Match` header and the `etag_of_fname` argument is provided, the middleware will check the `ETag` and respond with `304 Not Modified` if it matches. (Note that it will do this even if the file no longer exists in the filesystem).

This PR also adds an optional `headers` argument. Using this the user can e.g. specify a `Cache-Control` policy.

Aside: it looks like [`Cohttp_lwt_unix.Server.respond_file` now sets the `Content-Type` header using `Magic_mime`](https://github.com/mirage/ocaml-cohttp/blob/e295f75f249ffa51c8a368f2792872b58e4b2b99/cohttp-lwt-unix/src/server.ml#L40-L42). Should we remove that from this middleware?


rgrinberg:
> Aside: it looks like Cohttp_lwt_unix.Server.respond_file now sets the Content-Type header using Magic_mime. Should we remove that from this middleware?

Indeed.

Thanks for the PR.

Issue #73, opened 2017-12-05
Alternative mimetype lookup


malthe:
I need to return `application/javascript; charset=utf-8` instead of the default for `*.js` files.

(Since my source files include non-ascii characters.)

Should `Static_serve.m` take an optional mimetypes module that allows a custom lookup?

rgrinberg:
Sure, I think that would be useful. Why a module though? A function or some sort of map like structure isn't enough?

malthe:
A function would be good!

tmattio:
Fixed in #244

PR #78, opened 2018-07-04
Add mime_lookup optional param.

NightBlues:
Hi there! I'd like to contibute to opium a bit, so I found easy to implement issue (https://github.com/rgrinberg/opium/issues/73). Please consider my PR - I'm open to change it.


rgrinberg:
Thanks, but I actually think that this feature is better addressed via middleware. For example, you could easily create a middleware that will add the missing mime header

PR #79, opened 2018-07-11
Add lwt_log to dependencies

andrebauer:
Lwt_log is deleted from lwt >= 4.0.0, adding lwt_log to the dependencies allows opium to compile with lwt 4.1.0.


rgrinberg:
Thanks

PR #82, opened 2018-09-20
Expose all cookie directives

actionshrimp:
- Expose all the HTTP cookie directives, not just expiration
- Remove the defaults from `set_cookies` and use `Set_cookie_hdr.make`'s defaults. This is the same for the expiration (both are `Session`), but path is no longer fixed/defaulted to `/`.
- Use `Set_cookie_hdr.t` as the internal representation of the cookie used by the middleware as it has all the fields we want and a sexp representation.

NB: encoding of the cookie data is now done when adding to the Env rather than at the point of serialisation - this might actually make debugging a bit easier as it will be clearer that opium is encoding itself rather than something lower level? But happy to change this back if that seems sensible.

### DIFF
--- a/opium_kernel/cookie.ml
+++ b/opium_kernel/cookie.ml
@@ -53,11 +53,12 @@ let get req ~key =
     |> List.find_map ~f:(fun (k,v) ->
       if k = key then Some (decode v) else None)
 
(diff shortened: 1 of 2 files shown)


rgrinberg:
Thanks.

PR #83, opened 2018-09-20
String split fix for ocaml < 4.04

actionshrimp:
I noticed the travis build was failing on ocaml 4.03 due to `String.split_on_char` (introduced in `4.04`) . As `stringext` is already a transitive dependency I've added it explicitly and used that to keep compatibility with 4.03.


rgrinberg:
Thanks.

PR #85, opened 2018-10-02
Simplify module syntax

rvantonder:


rvantonder:
CI build unrelated: fails due to issue fixed in #83.

Issue #86, opened 2018-10-12
New release


XVilka:
Because of the Lwt update, basically current release fails to compile.

Schniz:
I guess you're talking about `Error: Unbound module Lwt_log`?

anuragsoni:
Not sure if you've noticed, but https://opam.ocaml.org/packages/opium/opium.0.17.0/ was released sometime ago.

rgrinberg:
@anuragsoni would you like to make a new release?

anuragsoni:
@rgrinberg Sure. I'll prepare one later today.

anuragsoni:
Closing as a new release has been published to opam now. https://opam.ocaml.org/packages/opium/opium.0.17.1/

PR #87, opened 2018-11-28
Switch to opam2 file format + dune

anuragsoni:
* Switch to dune and opam 2 file formats
* Try to remove usage of Lwt_log

Reference: #86


anuragsoni:
This is mostly just fixing opam lint errors, switching to dune files and using the built-in formatter and fixing any compilation errors after removing `lwt_log`.

I also bumped up the minimum ocaml version to match whatever `cohttp` sets at the moment.

Issue #89, opened 2019-01-17
How to set up logging?


lindig:
I am struggling to set up logging to syslog. My current attempt is something like this. It requires `logs-syslog.lwt` as an additional library. I'd appreciate any help.

```
open Opium.Std
open Lwt.Infix

let hello = get "/" (fun _ -> `String "Hello World" |> respond')

let _install_syslog () =
  let socket = "/var/run/syslog" in (* use /dev/log on Linux *)
  let prio   = Some Logs.Info in
  Logs_syslog_lwt.unix_reporter ~socket () >|= function
  | Ok r    -> Logs.set_reporter r; Logs.set_level prio
  | Error r -> failwith r

let () =
  let app =
    App.empty
    |> hello
    |> App.run_command' in
  match app with
  | `Ok app ->
    Lwt_main.at_exit (fun () -> Lwt.return (print_endline "Testing"));
    let s =
      _install_syslog () >>= fun () -> 
      Lwt.join [app; Lwt_unix.sleep 2.0 |> Lwt.map (fun _ -> Lwt.cancel app)] in
    ignore (Lwt_main.run s)
  | `Error -> exit 1
  | `Not_running -> exit 0
```

anuragsoni:
Continuing the discussion from discord. Here was a small example that seems to work for me.

```ocaml
open Opium.Std
open Lwt.Infix

let syslog_reporter () = Logs_syslog_lwt.unix_reporter ()

let say_hello =
  get "/hello/:name" (fun req ->
      `String ("Hello " ^ param req "name") |> respond' )

let _ =
  let a = App.empty |> say_hello |> App.run_command' in
  match a with
  | `Ok a ->
      let s =
        syslog_reporter ()
        >>= function
        | Ok r ->
            Logs.set_reporter r ;
            Logs.set_level (Some Logs.Debug) ;
            a
        | Error r -> failwith r
      in
      Lwt_main.run s
  | `Error -> exit 1
  | `Not_running -> exit 0
```

lindig:
Thank you. I can confirm that I now see log entries. (I'm on macOS where logging is a bit weird). I'm happy to see the ticket closed but would recommend to add this to the example section because it is a common problem.

anuragsoni:
Adding examples with logging would indeed be useful. I think there was at least one more issue #81  where there was confusion about logging not working.

[...]

PR #90, opened 2019-01-21
Logging example

anuragsoni:
Reference: #89

Signed-off-by: Anurag Soni <anurag@sonianurag.com>


anuragsoni:
I'd like to setup something that runs the examples in CI before i merge this.

PR #118, opened 2019-11-05
Note how to find the docs in the readme

shonfeder:
This should help with general accessibility. 

Please let me know if you'd prefer this addressed another, or feel free to reject and close if you'd rather do it yourself. But helping users find the docs should be implemented one way or the other! :)

### DIFF
--- a/README.cpp.md
+++ b/README.cpp.md
@@ -34,6 +34,16 @@ $ opam install opium
 $ opam pin add opium_core git+https://github.com/rgrinberg/opium.git
 $ opam pin add opium git+https://github.com/rgrinberg/opium.git
 ```
+
+## Documentation
+
+- Read [the hosted documentation for the latest version][hosted-docs].
+- Build and view the docs for version installed locally using [`odig`][odig]:
+  `odig doc opium`.
+
+[hosted-docs]: https://rgrinberg.github.io/opium/
+[odig]: https://github.com/b0-system/odig
+
 ## Examples
 
 All examples are built once the necessary dependencies are installed.
--- a/README.md
+++ b/README.md
@@ -34,6 +34,16 @@ $ opam install opium
 $ opam pin add opium_core git+https://github.com/rgrinberg/opium.git
 $ opam pin add opium git+https://github.com/rgrinberg/opium.git
 ```
+
+## Documentation
+
+- Read [the hosted documentation for the latest version][hosted-docs].
+- Build and view the docs for version installed locally using [`odig`][odig]:
+  `odig doc opium`.
+
+[hosted-docs]: https://rgrinberg.github.io/opium/
+[odig]: https://github.com/b0-system/odig
+
 ## Examples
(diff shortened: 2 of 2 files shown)


anuragsoni:
Thanks! We should indeed make it easy to find documentation :)

shonfeder:
Hi. I was looking things over gearing up to make some other small contributions, and it looks to me like this change has been dropped from `master` and `develop`. The commit is still visible on https://github.com/rgrinberg/opium/commits/experiment-with-splitting-modules 

Did the history of `master` get overwritten at some point? That's quite troubling if so...

anuragsoni:
I apologize, I must have accuse reverted this when I was undoing some changes about rewriting modules. I’ll bring this back to master when I’m near a computer later today.

anuragsoni:
For some background, I had some changes related to renaming opium_kernel to a core. I decided against going down that route. Now I’ll work on just getting opium_kernel to a state where can do most of the work without needing the dependency on cohttp.

roddyyaga:
There are a few other changes that are missing from master/develop too, e.g. the switch from ezjsonm to yojson

anuragsoni:
@shonfeder Your commit about the docs is back on master.

@roddyyaga The yojson switch was reverted for now as i'd like to re-evaluate that choice.

shonfeder:
Thanks, @anuragsoni 

Is the eventual aim with Opium to build out a more robust community project that is production ready?

If so, I suggest that we fix and codify the development process for this project so that we're not ever rewriting the history of master. From everything I've read and been taught about git best practices, there is virtually no reason we should ever be force pushing to master and rewriting history. If something has gone very wrong, and we need to revert changes to master, then new revert commits should be added so that the undo is recorded in the history. Moreover, if work is experimental and uncertain, then it shouldn't be in master in the first place, but in an experimental feature branch.

Let me know if you'd like to talk this over at all, or would like any help in coming up with a development process that will avoid this kind of dangerous and disorienting problem in the future.

Thanks again for the fix and for your work maintaining the library!

anuragsoni:
@shonfeder I really appreciate the concern and I’d welcome more people contributing, and I’m very willing to have a contribution guideline setup. 

That being said, this particular situation isn’t normal and it isn’t what would happen during regular development. As far as I was aware there wasn’t anyone else contributing to the library. There were things that were easy to do the way I did, to get things to a place where it’s easy to have a clean slate. This one change got lost during that and that’s a mistake on my part. 

I am happy you decided to contribute and I’ll look forward to future contribution and feedback to help improve opium.

shonfeder:
Thanks @anuragsoni. I appreciate you taking my concerns into account.

anuragsoni:
@shonfeder not at all. It’s always good to hear from actual users and it’s good feedback. Note that the next big change will just be to eventually move to httpaf instead of cohttp. Some other low hanging changes can be done without disrupting the public API.

PR #126, opened 2019-12-30
Add documentation for urlencoded_pairs_of_body

shonfeder:
Small documentation add. The main intent is to point users in the right place when they search for "form data".


anuragsoni:
Thank you!

PR #149, opened 2020-06-01
Kernel middlewares

tmattio:
Hi,

As we discussed, this PR lays the foundation to provide middlewares directly in Opium (or rather, in `opium_kernel`).

It does not add any new middleware, but moves the existing ones to `opium_kernel` and improves them a bit.

Since it's a general PR, I also committed a few changes to the project setup (adding commands to `Makefile`, `odoc` formatting in `.ocamlformat`, etc.). I can revert those if you prefer to keep the setup as-is.

Another change is the addition of a `Halt` exception in `Rock` that can be used to prematurely respond to the client, and thus by-pass any other middleware of the pipeline. This is useful in the context of an error handler middleware, or in our case, the debugger middleware.

## Logger Middleware

The logger middleware (previously trace), logs the requests and responses. Since CoHTTP used to log the body of the request/response in Debug verbosity, it reproduces this behavior.

It can also time the handling of the response if a `time_f` function is passed. The middleware defined in `Opium.Middleware.logger` wraps `Opium_kernel.Middleware.logger` and provide a `time_f` function that uses `Unix.gettimeofday` to time the response handling.

## Debugger Middleware

The debugger middleware (previously debug) catches any exception raised during the handling of the response and generates an HTML page with the pretty-printed request and the pretty-printed exception.

The only change to the logic is the addition of the halt exception that will cause the response to be sent directly if an exception is caused, instead of passing the response through all of the other middlewares.

I also spent some time making the error page "good-looking". Here's how it looks like:

image

## Todo

As part of this PR, I'd also like to implement:

- [x] Better pretty printer for requests
- [x] An error handler middleware

But let me know if you prefer that I open other PRs for these.

EDIT:

- I reverted the Halt exception to handle early responses and opened a different PR for it


TODO.md
```diff
@@ -1,9 +1,7 @@
 ## Things missing when compared to version based on Cohttp
 
-* ~~Request body isn't handled yet (most important task for now)~~
-* ~~No sexp derivers (Httpaf doesn't have sexp derivers for their types, consider using their pretty printers instead?)~~
-* ~~No static file serving.~~
```
tmattio, line 5: I assumed these have been taken care of. Is it the case?

opium/app.ml
```diff
@@ -83,8 +83,8 @@ let attach_middleware { verbose; debug; routes; middlewares; _ } =
   in
   [ Some (routes |> create_router |> Router.m) ]
   @ ListLabels.map ~f:Option.some middlewares
-  @ [ (if verbose then Some Debug.trace else None)
-    ; (if debug then Some Debug.debug else None)
+  @ [ (if verbose then Some Logger.m else None)
+    ; (if debug then Some (Opium_kernel.Middlewares.debugger ()) else None)
```
tmattio, line 87: We probably want to include `Opium_kernel.Middleware` in `Opium.Middleware` to avoid having to have a single entry point when building apps with `opium`.

opium_kernel/opium_kernel.mli
```diff
@@ -190,3 +188,16 @@ module Server_connection : sig
     -> Rock.App.t
     -> 'a Lwt.t
 end
+
+module Middlewares : sig
+  val router
+    :  (Route.t * (Rock.Request.t -> Rock.Response.t Lwt.t)) Queue.t array
```
tmattio, line 194: I can simplify the types here.

tmattio:
I added error handler middlewares as part of this PR.

There are two error handlers: `Middleware.json_error_handler` and `Middleware.html_error_handler`.

The former returns a JSON object with the following fields:

- status
- error
- message

Here's an example for an internal server error:

```
{"status":500,"error":"Internal Server Error","message":"The server encountered an unexpected condition that prevented it from fulfilling the request."}
```

The later returns an HTML page with the same information. Here's an example with the same error (internal server error):

image

opium_kernel/middlewares/error_handler.ml
```diff
@@ -0,0 +1,270 @@
...
+end
+
+module Json = struct
+  module Error = struct
+    type t =
+      { status : int
+      ; error : string
+      ; message : string
+      }
+    [@@deriving yojson]
```
anuragsoni, line 258:
```suggestion
    let to_yojson { status; error; message } =
      `Assoc [ "status", `Int status; "error", `String error; "message", `String message ]
    ;;
```

The struct is small enough that creating a yojson type manually is simple. I'd prefer that instead of pulling in a ppx dependency. The removal of all ppx deps was a conscious decision and i don't think we should add a ppx deps unless we benefit a lot from using it.

opium_kernel/middlewares/logger.ml
```diff
@@ -0,0 +1,66 @@
...
+
+let log_src = Logs.Src.create "opium.server"
+
+let log_body Core.Body.{ content; _ } =
+  match content with
+  | `Stream _ -> ()
+  | `Empty -> ()
+  | `String s -> Logs.debug ~src:log_src (fun m -> m "%s" s)
+  | `Bigstring b -> Logs.debug ~src:log_src (fun m -> m "%s" (Bigstringaf.to_string b))
+;;
```
anuragsoni, line 28: This could be simplified by using the existing pretty printers for the Body type

```suggestion
let log_body body = Logs.debug ~src:log_src (fun m -> m "%a" Core.Body.pp_hum body)
```

opium_kernel/middlewares/error_handler.ml
```diff
@@ -0,0 +1,270 @@
+(** [Error_handler] creates error HTML pages for standard HTTP errors.
```
anuragsoni, line 1: I'm not entirely sure about this middleware. The change about the detailed default message looks good. But i don't think this needs to be a middleware. We already have a error handler defined at

https://github.com/rgrinberg/opium/blob/c2db89ffc317fb5d6dd2283c198b198d94e830fd/opium_kernel/server_connection.ml#L3

This can be overriden by users via an input to the Server_connection. The user needs to provide a handler with the following type

https://github.com/rgrinberg/opium/blob/c2db89ffc317fb5d6dd2283c198b198d94e830fd/opium_kernel/opium_kernel.mli#L180-L183

The httpaf runtime will watch for any exception in the handlers and wake up the error_handler when that happens. So that seems like a natural place to deal with errors that occur via an unrecoverable exception. For opium i lean towards providing a decent default httpaf error handler, (while allowing user to define their own behavior if desired), to deal with the exception scenarios. For application level errors i think it should be up to the user to create a middleware that's specific for their own application that can fill this role.

opium_kernel/core/rock.ml
```diff
@@ -153,6 +205,16 @@ module Response = struct
         ])
   ;;
 
+  let string_of_t t =
```
anuragsoni, line 208: Is this done just to avoid s-exp formatting? I had added pretty printers for all the with this usecase in mind. Getting a string from the formatter can be done via `Format.asprintf`

opium_kernel/core/rock.ml
```diff
@@ -88,6 +126,16 @@ module Request = struct
...
```
anuragsoni, line 129: same comment about the existing pretty printer via Format.asprintf

opium_kernel/core/rock.ml
```diff
@@ -27,13 +27,23 @@ module Pp = struct
         ])
```
anuragsoni, line 27: General comment about the diff for this file. I'd lean towards using the existing pretty printers (pp_hum) functions instead of something that just converts things to a string. The formatters are more flexible and still allow converting content into a string via `Format.asprintf "%a" M.pp_hum`

opium_kernel/core/rock.ml
tmattio: I added these functions to improve the formatting. I find the s-exp not very readable.

Here's an extract of the logs with the s-exp:
```
middleware_debugger.exe: [DEBUG] ((version (major 1) (minor 1)) (status 500)
                                  (reason ()) (headers ())
                                  (body
                                   ((length (6810))
                                    (content
                                      "\
                                     \n<!doctype html>\
                                     \n<html lang=\"en\">\
                                     \n\
                                     \n<head>\
```

Compared to the logs with an HTTP format:

```
middleware_debugger.exe: [DEBUG] HTTP/1.1 500 



<!doctype html>
<html lang="en">

<head>
```

Same goes for the debugger.

Screenshot 2020-06-02 at 11 29 34

Compared with:

Screenshot 2020-06-02 at 11 29 14


But I agree about the flexibility. I added a formatted called `pp_http` in both Request and Response and use them in the logger and debugger.

tmattio:
Thanks a lot for the review!

I addressed your comments and I've pull out the error handler (will create another PR).
The only thing left is the formatting of requests/responses in logger and debugger, (see my answer to your comment) 🙂

opium_kernel/middlewares/error_handler.ml
tmattio: I've pulled out the error handler from this PR and I gave a bit more context in https://github.com/rgrinberg/opium/pull/156

anuragsoni:
Thank you!

PR #151, opened 2020-06-01
Add a Halt exception to allow early responses

tmattio:
As discussed, I extracted the early response part of #149 into a single PR.

The reason for adding this is that some middleware might want to take control of the request flow and decide to return the response early. This is the case for instance for an error handler, where once the error is caught and the response is generated, we most likely do not want other middleware to mess with the response.

The implementation is inspired by https://github.com/ocamllabs/ocaml-effects-tutorial.
I've been trying to think of other solutions, but I can't think of one that wouldn't add a lot of overhead. I'm happy to try other implementations if you have any idea 🙂

### DIFF
--- a/opium_kernel/src/debugger.ml
+++ b/opium_kernel/src/debugger.ml
@@ -95,7 +95,7 @@ let m () =
       (fun exn ->
         Logs.err ~src:log_src (fun f -> f "%s" (Nifty.Exn.to_string exn));
         let body = format_error req exn |> Body.of_string in
-        Lwt.return @@ Response.make ~status:`Internal_server_error ~body ())
+        halt (Response.make ~status:`Internal_server_error ~body ()))
   in
   Middleware.create ~name:"Debugger" ~filter
 ;;
--- a/opium_kernel/src/opium_kernel.mli
+++ b/opium_kernel/src/opium_kernel.mli
@@ -174,6 +174,20 @@ module Rock : sig
     val append_middleware : t -> Middleware.t -> t
     val create : ?middlewares:Middleware.t list -> handler:Handler.t -> unit -> t
   end
+
+  (** The Halt exception can be raised to stop the interrupt the normal processing flow of
+      a request.
+
+      The exception will be handled by the main run function (in {!Server_connection.run})
+      and the response will be sent to the client directly.
+
+      This is especially useful when you want to make sure that no other middleware will
+      be able to modify the response. *)
+  exception Halt of Response.t
+
+  (** Raises a Halt exception to interrupt the processing of the connection and trigger an
+      early response. *)
+  val halt : Response.t -> unit
 end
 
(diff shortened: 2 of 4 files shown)


anuragsoni:
Hmm, apart from the name, this is possible to achieve with the current implementation as well.

```ocaml
exception Early_exit

let reject_sometimes =
  let filter handler req =
    let target = req.Request.target in
    let query_params = Uri.query (Uri.of_string target) in
    let p = List.assoc_opt "action" query_params in
    match p with
    | Some [ "reject" ] ->
      raise_notrace Early_exit
    | _ -> handler req
  in
  Rock.Middleware.create ~filter ~name:"reject sometimes"
;;
```

>  but I can't think of one that wouldn't add a lot of overhead.
I'm not sure what overhead you had in mind, but typically if i'm thinking of a chain of monads where i need an early exit, i reach for `Result.t`, or `Option.t`. That could be another alternative that can work here if the service was to change from the current implementation to be:

```ocaml
type service = Request.t -> (Response.t, error) Lwt_result.t
```

That said, i'm not sure i can think of many cases where such logic lives inside a middleware, or middlewares that do early exits in success cases.

tmattio:
> Hmm, apart from the name, this is possible to achieve with the current implementation as well.

Without any change, I think this example will throw an exception and will result in a 500. To implement the early response behavior, the exception would have to take a response that will be handled in the Server_connection. Unless I'm missing something?

> if i'm thinking of a chain of monads where i need an early exit, i reach for Result.t, or Option.t

Yup, that's the overhead I had in mind. The early response is a very specific use case and it seems like an unnecessary overhead to ask users to wrap their response in yet another monad in all of their middlewares. I think it also makes the middleware interface less intuitive. The Option monad is not usable here, because we need the response in any case. As for the Result monad, users would have to return the response as an error, which intuitively does not translate well to "this error response will be returned without going through the other middlewares".

rgrinberg:
The idea looks fine to me, but it needs to live in a middleware. I wouldn't call it an algebraic effect either, as users's can't really install their own handler. Most users are unlikely to even know what an algebraic effect is for that matter.

As for monads, the less monads the better. OCaml's support for composing is pretty poor and once we have actual algebraic effects, they will be mostly useless for concurrency as well.

tmattio:
@rgrinberg I'm going to update the documentation to make clearer and add some examples.

I'm not sure which part you would put in a middleware? If you're talking about the exception handling, the problem is, it has to be handled in the `run` function to prevent other middleware to be processed. If we put the exception handling in a middleware, it wouldn't raise anymore, but other middlewares would still process the request. Maybe there's a solution I'm not seeing?

tmattio:
I updated the documentation to remove the fancy terminology. Going to merge now, but let me know if you still have comments, I'll still be able to address them in other PRs.

Issue #96, opened 2019-08-13
How can we override default 500, 404 page ?


Et7f3:
So I could send for instance link to the api documentation for instance. Or styling my 404 pages.

anuragsoni:
We'll need to allow overriding the default `not_found` handler so it'll be possible to override the default 404 page. I believe the 500 message comes from Cohttp

Et7f3:
If they occur when a execptio is launched opium could try ... with all exception then ? Or we should open issue on Cohttp and maybe httpaf (when opium will change his backend)

Et7f3:
Maybe add a try_all bool somewhere then that is `false` by default. Maybe inside App.empty

PR #156, opened 2020-06-02
Error handler middleware

tmattio:
This provides error handler middlewares for JSON and HTML.

There are mainly three reasons I implemented this as a middleware.

- There was already a default HTML page for the error 404, defined in `Opium.App`. So this middleware can be seen as a generalization of this error page to every error code.

- The error handler defined in `Server_connection` will only work when exceptions are raised.
  The exception does not contain the error code, and reporting errors other than exceptions are reserved to Httpaf internal uses (functions are not exposed)
  so if users want to rely on the Server_connection error handler, they are limited to using the internal server error, which seems pretty limited.

  I think most of the time, users will return an HTTP response with an error code instead of raising an exception. If there is no error handler, they will have to write their own.
  Most applications will end up having to do this at some point, but following the design goal "A programmer should be instantly productive when starting out.", I thought it would be nice to provide good defaults.

- A lot of applications have different types of entry points. It is common to provide JSON endpoints as well as HTML endpoints in the same application.
  The error handlers will be different depending on the content type of the endpoint, and a middleware provides this flexibility as it can be applied to specific routes.

I also provided a way to easily override the error handling of specific status code with https://github.com/rgrinberg/opium/issues/96 in mind. One can override an error with the following syntax:

```
let custom_handler = function
  | `Forbidden -> Some (Response.of_string "Denied!")
  | _ -> None
;;

let error_handler = Middleware.error_handler ~custom_handler ()
```

which, IMHO, is a bit more user friendly than implementing a custom Httpaf error handler.

Error handling is an important part of a web framework and I think it's crucial that we find a good balance between convention, productivity, and flexibility. So I'm happy to experiment on other solutions to provide good defaults if the middleware solution seems weird to you.


tmattio:
Closing this as it seems there are better ways to manage errors than a middleware.

PR #162, opened 2020-06-05
Add Httpaf modules

tmattio:
As there are more and more helper functions for the Httpaf types, I thought it would be good to put them in their own module.

It also makes the API more consistent: users can use `Opium_kernel` all the time instead of switching between `Opium_kernel` and `Httpaf` depending on the function they are looking for.

I also added a few helper functions in `Rock` that I used when implementing some middlewares.


opium_kernel/src/core/status.ml
```diff
@@ -0,0 +1,11 @@
+include Httpaf.Status
+
+let sexp_of_t status =
+  let open Sexplib0 in
+  let open Sexp_conv in
+  sexp_of_int (to_code status)
```
anuragsoni, line 6:
```suggestion
  Sexplib0.Sexp_conv.(sexp_of_int (to_code status))
```

opium_kernel/src/middlewares/logger.ml
```diff
@@ -29,29 +29,30 @@ let respond ?time_f handler req =
     let span, response_lwt = time_f f in
     response_lwt
     >|= fun response ->
-    let code = response.Response.status |> Httpaf.Status.to_string in
-    Log.info (fun m -> m "Responded with HTTP code %s in %a" code Mtime.Span.pp span);
-    Log.debug (fun m -> m "%a" Response.pp_http response);
+    let code = response.Response.status |> Status.to_string in
+    Logs.info ~src:log_src (fun m ->
```
anuragsoni, line 33: This change from `Log.info` to `Logs.info ~src:log_src` is not required. module Log is defined as

```ocaml
let log_src = Logs.Src.create "opium.server"

module Log = (val Logs.src_log log_src : Logs.LOG)
```
All logging functions in Log are for the source log_src

opium_kernel/src/middlewares/logger.ml
```diff
@@ -29,29 +29,30 @@ let respond ?time_f handler req =
...
+    Logs.debug ~src:log_src (fun m -> m "%a" Response.pp_http response);
     response
   | None ->
     handler req
     >|= fun response ->
-    let code = response.Response.status |> Httpaf.Status.to_string in
-    Log.info (fun m -> m "Responded with HTTP code %s" code);
-    Log.debug (fun m -> m "%a" Response.pp_http response);
+    let code = response.Response.status |> Status.to_string in
+    Logs.info ~src:log_src (fun m -> m "Responded with HTTP code %s" code);
```
anuragsoni, line 41: Same comment about Log <-> Logs here

opium_kernel/src/opium_kernel.mli
```diff
@@ -32,6 +32,22 @@ module Body : sig
   val of_stream : ?length:Int64.t -> string Lwt_stream.t -> t
 end
 
+module Headers : sig
+  include module type of Core.Headers
+end
```
anuragsoni, line 37: I think this can be written as:

```suggestion
module Headers : module type of Core.Headers
```

opium_kernel/src/opium_kernel.mli
```diff
@@ -32,6 +32,22 @@ module Body : sig
...
+
+module Version : sig
+  include module type of Core.Version
+end
+
+module Status : sig
+  include module type of Core.Status
+end
+
```
anuragsoni, line 50: Same comment as for the `module Headers : module type of Core.Headers`

PR #163, opened 2020-06-05
Add CORS middleware

tmattio:
This PR adds a middleware to manage CORS headers.

It adds the appropriate headers for regular requests and handles [preflight requests](https://developer.mozilla.org/en-US/docs/Glossary/preflight_request).

It depends on #162, but I can refactor if you prefer not to merge it.

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -39,7 +39,8 @@
   bigstringaf
   re
   uri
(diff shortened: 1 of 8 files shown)


anuragsoni:
Thank you!

PR #172, opened 2020-07-03
Allow using a custom host to bind the server.

rizo:
Currently the server binds to `inet_addr_loopback` (_i.e._ localhost). This PR adds support for setting a custom host when starting the app.

The host can be set similarly to the port with a builder option:

```ocaml
  App.empty
  |> App.host "localhost"
  |> App.port port
  ...
  |> App.start
```

By default the host is set to `0.0.0.0`.


anuragsoni:
Thanks!

PR #179, opened 2020-08-30
Add helper functions to Rock's Request and Response modules

tmattio:
## Proposed Changes

This PR adds numerous helper functions to Opium Kernel, most of them in the Request and Response modules. The list of new functions comes from my own needs while working on a web application with Opium, so hopefully, this should be close to what other users will need too.

While adding new functionalities to the Request and Response modules, I tried to build a cohesive API. I noticed that a lot of helper functions were spread over different modules (e.g. the `param` function in `Router`, or the `json` function in `App`), so I moved these functions to their respective modules. After this PR, all of the helper functions should be located in either Request or Response, which I think offers a much clearer API, especially for beginners.

Lastly, I moved the Request, Response, and the other HTTP-related modules to the top-level. With the addition of the helper functions, I think the Request and Response modules are important enough that they need to be exposed at the top level in the documentation. With this change, it was natural to move the other HTTP-related modules to the top-level, for consistency.

This introduces breaking changes. The reasoning is that we already paid the cost of introducing huge breaking changes to the API with the switch to Httpaf. The API was not compatible anymore, and the modules Headers, Status, and co. didn't even exist before. Secondly, I think the cost of breaking additional API that existed before the switch to Httpaf, such as the `param` function in `Router` is acceptable given the improvements on the overall API

## Documentation

I focused on the documentation for the `Request` and `Response` modules. This is part of a bigger initiative to improve Opium's documentation, but the output is representative of what I imagine for the documentation of each module.

I separated every function into sections and documented each of them with a code example of how to use the function.

I also added a sub-section on top of each function to display them on the sidebar. Here's what the documentation looks like for the Request module:

image

### `Opium_kernel.Rock`

- 🟠 `Opium_kernel.Rock.Request` moved to `Opium_kernel.Request`
- 🟠 `Opium_kernel.Rock.Response` moved to `Opium_kernel.Response`
- 🟠 `Opium_kernel.Rock.Body` moved to `Opium_kernel.Body`
- 🟠 `Opium_kernel.Rock.Headers` moved to `Opium_kernel.Headers`
- 🟠 `Opium_kernel.Rock.Method` moved to `Opium_kernel.Method`
- 🟠 `Opium_kernel.Rock.Version` moved to `Opium_kernel.Version`
- 🟠 `Opium_kernel.Rock.Status` moved to `Opium_kernel.Status`

### `Opium.App`

- 🟠`Opium.App.param` was removed in favor of `Opium_kernel.Router.param`
- 🟠`Opium.App.splat` was removed in favor of `Opium_kernel.Router.splat`
- 🟠 `Opium.App.string_of_body_exn` moved to `Response.of_plain_text`
- 🟠 `Opium.App.urlencoded_pairs_of_body` moved to `Response.of_urlencoded`

### `Opium_kernel.Body`

- 🟢`Opium_kernel.Body.copy` was added
- 🟢`Opium_kernel.Body.pp` was added
- 🟢`Opium_kernel.Body.pp_hum` was added
- 🟠`Opium_kernel.Body.t` is now a public type
- 🟠`Opium_kernel.Body.content` is now a public type

### `Opium_kernel.Request`

- 🟢`Opium_kernel.Request.of_plain_text` was added
- 🟢`Opium_kernel.Request.of_json` was added
- 🟢`Opium_kernel.Request.of_urlencoded` was added
- 🟢`Opium_kernel.Request.to_json_exn` was added
- 🟢`Opium_kernel.Request.to_json` was added
- 🟢`Opium_kernel.Request.to_plain_text` was added
- 🟢`Opium_kernel.Request.to_urlencoded` was added
- 🟢`Opium_kernel.Request.header` was added
- 🟢`Opium_kernel.Request.headers` was added
- 🟢`Opium_kernel.Request.add_header` was added
- 🟢`Opium_kernel.Request.add_header_or_replace` was added
- 🟢`Opium_kernel.Request.add_header_unless_exists` was added
- 🟢`Opium_kernel.Request.add_headers` was added
- 🟢`Opium_kernel.Request.add_headers_or_replace` was added
- 🟢`Opium_kernel.Request.add_headers_unless_exists` was added
- 🟢`Opium_kernel.Request.remove_header` was added
- 🟢`Opium_kernel.Request.content_type` was added
- 🟢`Opium_kernel.Request.set_content_type` was added
- 🟢`Opium_kernel.Request.urlencoded` was added
- 🟢`Opium_kernel.Request.urlencoded_exn` was added
- 🟢`Opium_kernel.Request.query_list` was added
- 🟢`Opium_kernel.Request.query` was added
- 🟢`Opium_kernel.Request.query_exn` was added

### `Opium_kernel.Response`

- 🟢`Opium_kernel.Response.redirect_to` was added
- 🟠 `Opium_kernel.Response.of_string` moved to `Opium_kernel.Response.of_plain_text`
- 🟢`Opium_kernel.Response.of_html` was added
- 🟢`Opium_kernel.Response.of_xml` was added
- 🟢`Opium_kernel.Response.of_svg` was added
- 🟢`Opium_kernel.Response.header` was added
- 🟢`Opium_kernel.Response.headers` was added
- 🟢`Opium_kernel.Response.add_header` was added
- 🟢`Opium_kernel.Response.add_header_or_replace` was added
- 🟢`Opium_kernel.Response.add_header_unless_exists` was added
- 🟢`Opium_kernel.Response.add_headers` was added
- 🟢`Opium_kernel.Response.add_headers_or_replace` was added
- 🟢`Opium_kernel.Response.add_headers_unless_exists` was added
- 🟢`Opium_kernel.Response.remove_header` was added
- 🟢`Opium_kernel.Response.content_type` was added
- 🟢`Opium_kernel.Response.set_content_type` was added
- 🟢`Opium_kernel.Response.etag` was added
- 🟢`Opium_kernel.Response.set_etag` was added
- 🟢`Opium_kernel.Response.location` was added
- 🟢`Opium_kernel.Response.set_location` was added
- 🟢`Opium_kernel.Response.cache_control` was added
- 🟢`Opium_kernel.Response.set_cache_control` was added
- 🟢`Opium_kernel.Response.to_json_exn` was added
- 🟢`Opium_kernel.Response.to_json` was added
- 🟢`Opium_kernel.Response.to_plain_text` was added

## Todo

List of things to do before this is ready:

- [x] Complete the documentation
- [x] Make a list of API changes
- [ ] Decide on what to do for the Body types (they are not private anymore because the `mli` is not inlined, should we make them abstract?)


tmattio:
@rgrinberg @anuragsoni Tagging you here since this is introducing breaking changes. I intended for this PR to be additive only, but after seeing an opportunity to make the API clearer by aggregating the helper functions in `Request` and `Response`, this turned into a refactoring of several functions. There's no big structural refactoring (apart for the move of the HTTP modules to the top-level), so I guess it boils down to: does it make sense to you to have all of the helper functions in the `Request` and `Response` modules?

dune-project
```diff
@@ -36,9 +36,11 @@
   logs
   mtime
   yojson
+  tyxml
```
tmattio, line 39: This might be controversial. The addition of Tyxml as a dependency allows Opium to provide several helper functions (`Response.of_html`, `Response.of_svg`). I figured these are common enough that it justifies the addition of a new dependency.

examples/hello_world.ml
```diff
@@ -11,16 +11,15 @@ end
...
+      let name = Request.param "name" req |> Option.get in
+      let age = Request.param "age" req |> Option.get |> int_of_string in
+      let person = { Person.name; age } |> Person.yojson_of_t in
       Lwt.return (Response.of_json person))
 ;;
 
 let update_person =
   patch "/person" (fun req ->
-      let+ json = App.json_of_body_exn req in
+      let+ json = Request.json req in
```
tmattio, line 22: This should be `json_exn`

opium_kernel/src/body.ml
```diff
@@ -40,16 +40,6 @@ let sexp_of_t { length; content } =
       ])
 ;;
 
-let string_of_t t =
```
tmattio, line 43: Deleted in favor of a pretty-printer, to have the same pair of functions `pp` and `pp_hum` as the other modules

opium_kernel/test/dune
```diff
@@ -1,4 +1,4 @@
 (tests
- (names core_route middleware_allow_cors)
```
tmattio, line 2: This was named core_route to reflect a directory structure that does not exist anymore.

dune-project
anuragsoni: Not controversial at all. Tyxml is pretty popular and the install counts on opam justify adding helpers for tyxml in my opinion :)

tmattio:
@anuragsoni The only change I am not sure about is the removal of the `private` types for the Body module. As you mentioned in #180, this might not be a good solution.

However I'd like to finish adding some middleware and a cookie module in the next week, so I wouldn't like this discussion to block everything. If that's ok with you, I'd like to go forward with this nonetheless and continue the discussion in #180. Once we come up with a good solution, I'll make the appropriate changes. What do you think?

anuragsoni:
> However I'd like to finish adding some middlewares and a cookie module in the next week and this depends on this PR, so I wouldn't like this discussion to block everything. If that's ok with you, I'd like to go forward with this nonetheless and continue the discussion in #180. Once we come up with a good solution, I'll make the appropriate changes. What do you think?

Of-course. What do you think about just making the types abstract for now. There are functions in the public api of body already that allow a user to get content out of it in their preferred format. I don't think there will be a feature loss if we go down that route.

tmattio:
> What do you think about just making the types abstract for now

That's what I was thinking also, but since the module definitions are inlined now (to avoid the `This expression has type Body.t but an expression was expected of type
         Opium_kernel__Body.t` compilation error), other modules in Opium_kernel need the types to be public, as they pattern match against the content (e.g. in `Server_connection.run`). I guess we could implement functions to address these, but that smells a bit hacky.

tmattio:
I made the types abstract in this commit : https://github.com/rgrinberg/opium/pull/179/commits/5ecbd8ffda9b33d0ef2e14fc400c2b0b9cbcb932

I brought back the module definition in `opium_kernel.mli` and added a `with type t = Body.t` at the end to solve the error I mentioned. But I'm really not sure if this is a good practice 🤔

tmattio:
Merging as is, but as I mentioned, let's keep the discussion going and I can revisit this part once we have a better solution 🙂

anuragsoni:
>  other modules in Opium_kernel need the types to be public, as they pattern match against the content (e.g. in Server_connection.run)
> I brought back the module definition in `opium_kernel.mli` and added a `with type t = Body.t` at the end to solve the error I mentioned. But I'm really not sure if this is a good practice 🤔

The signatures for all the modules were in the opium_kernel.mli for precisely this reason :) Having that content in the opium_kernel.mli meant that modules inside opium_kernel can still use pattern matching to access content from the body, and we still ensure that the types stay private/abstract for external libraries. I felt that approach was pretty nice to allow for ease of access to the types within the library but still ensuring some api boundary for modules not part of opium_kernel.

tmattio:
I liked it also, but with the documentation of the Request and Response module (and now middlewares), the file became hard to navigate (1000+ LOC) and I figured it didn't scale very well to have all the documentation centralized in one place.

Issue #176, opened 2020-07-31
Changes in API from Cohttp version


roddyyaga:
In the Cohttp version, there is a `respond` function and `body` type like this:

```ocaml
let to_body = function
  | `String s -> Body.of_string s
  | `Empty -> Body.empty
  | `Json j -> j |> Yojson.Safe.to_string |> Body.of_string

let respond ?headers ?status body =
  Response.make ?headers ?status ~body:(to_body body) () |> Lwt.return
```

Personally I thought that was a nice API, is there a reason it's been removed from the latest version?

anuragsoni:
Hi @roddyyaga ,

There are indeed some differences in the API compared to the cohttp version. One my reasoning for some changes was to move more of the response/body creation into opium_kernel (as opposed to opium) and have something that's easier to discover than poly variants. The current state of Body and Response look like this:

Body:
https://github.com/rgrinberg/opium/blob/29165400123c2d54da27d3e6edb3404374f532f9/opium_kernel/src/opium_kernel.mli#L41-L58

Response:
https://github.com/rgrinberg/opium/blob/29165400123c2d54da27d3e6edb3404374f532f9/opium_kernel/src/opium_kernel.mli#L121-L147

I felt that this would be easier for new users to discover than poly variants since this will be inside modules that they'd expect them to be in. I can see expanding the Response module to cover the other body types (stream is the only one that remains) which should bring this to par with the cohttp version.

Best,
Anurag Soni


On Fri, Jul 31, 2020, at 6:10 AM, roddyyaga wrote:
> 

> In the Cohttp version, there is a `respond` function and `body` type like this:

> let to_body = function
  | `String s -> Body.of_string s
  | `Empty -> Body.empty
  | `Json j -> j |> Yojson.Safe.to_string |> Body.of_string

> let respond ?headers ?status body =
  Response.make ?headers ?status ~body:(to_body body) () |> Lwt.return

PR #180, opened 2020-08-30
Make Rock.Body types public

tmattio:
Having the Body types private prevents users from implementing middlewares that make use of the body of the request. For instance, it is currently difficult to implement a custom logger.

@anuragsoni Is there any reason for making these types private? I don't want to undo a change that you made for a specific reason.

Related to https://github.com/rgrinberg/opium/issues/176

### DIFF
--- a/opium_kernel/src/opium_kernel.mli
+++ b/opium_kernel/src/opium_kernel.mli
@@ -21,7 +21,6 @@ module Rock : sig
 
   module Body : sig
     type content =
-      private
       [ `Empty
       | `String of string
       | `Bigstring of Bigstringaf.t
@@ -29,7 +28,7 @@ module Rock : sig
       ]
 
     (** [t] represents an HTTP message body. *)
-    type t = private
+    type t =
       { length : Int64.t option
       ; content : content
       }


anuragsoni:
> Having the Body types private prevents users from implementing middlewares that make use of the body of the request. For instance, it is currently difficult to implement a custom logger.

I might be missing something here. What kind of middlewares would a a user not be able to implement with the current public mli? The user has access to getting content out of the body via `to_string`, `to_stream` etc. The example shared on #176 was about a poly variant that was part of Opium, not the Body type that was used internally. Maybe bringing back a compat module that retains the opium api from 0.18 release is worth it? 

> Is there any reason for making these types private? I don't want to undo a change that you made for a specific reason.

Looking at the mli again, i think instead of making the types private i should've made them abstract and added a `val length : t -> int64`. I am not sure we need to expose the internals of how the body type is represented for the module to be usable. 
The intention with making the type private to allow for pattern matching on the body to extract the length and content, while ensuring things get created via the public api.

tmattio:
#179 has been merged and made the Body types abstract. I'll open an issue to continue the discussion with a better description of the problem / use cases.

PR #181, opened 2020-08-30
Fix make utop alias

tmattio:


Makefile
```diff
@@ -41,7 +41,7 @@ watch:
 
 .PHONY: utop
 utop:
-	opam exec -- dune utop lib -- -implicit-bindings
+	opam exec -- dune utop opium -- -implicit-bindings
```
rgrinberg, line 44: Perhaps `$ dune utop .` is even simpler.

PR #152, opened 2020-06-01
Static middleware

tmattio:
This PR implements a static serve middleware in `opium_kernel`. The implementation is based on the previous `Opium.Static_serve.m` middleware, but takes a function `read : string -> string option Lwt.t`.

This allows using the middleware with different `read` implementation (e.g. MirageOS's filesystem, ocaml-crunch, etc.)

I refactored the existing `Static_serve.m` middleware to use `Opium_kernel.Middleware.static`. The API is the same as the previous implementation.

PR #182, opened 2020-09-06
Add magic mime opium kernel

anuragsoni:
#152 moved `magic-mime` dependency to `opium_kernel`, but the opam files didn't reflect this. This resulted in a compilation error when pinning `opium_kernel` because of the missing dependency.


tmattio:
Sorry about that! Thanks for the fix 🙂

rgrinberg:
It will be good to switch to dune's opam file generation as well

anuragsoni:
> It will be good to switch to dune's opam file generation as well

@rgrinberg We are already using dune's opam file generation :)

PR #183, opened 2020-09-08
Improve CI/CD

tmattio:
This PR improves the CI CD by:

- Caching Opam dependencies
- Installing dependencies from lock files, which will make the builds more reproducible
- Splitting the steps and giving them names to identify failures quickly


rgrinberg (approved):

anuragsoni:
@tmattio restoring cache might cause issues with a later switch creation. Right now the ocaml action uses `opam switch create "$1" "$1"` to create a switch. I think that'll fail if there's an existing switch with the same name already present.

https://github.com/avsm/setup-ocaml/blob/4b93f69dd17a2b0bd35732d2573e57e307085f1e/lib/install-ocaml-unix.sh#L11

tmattio:
@anuragsoni That's what I noticed after merging. I'll comment out the caching part until there is a workaround. That's a shame, the CI went from taking 20min to 2min!

.github/workflows/ci.yml
```diff
@@ -0,0 +1,59 @@
...
+        with:
+          ocaml-version: ${{ matrix.ocaml-version }}
+
+      - name: Install dependencies
+        if: steps.cache-opam.outputs.cache-hit != 'true'
+        run: |
+          opam install . --deps-only --with-doc --with-test --locked --unlock-base
+          opam install ppx_sexp_conv base64 ppx_yojson_conv --skip-updates
+
+      - name: Upgrade dependencies
```
smorimoto, line 48: This is NOT for upgrade. This step checks the consistency of the installed opam packages.

.github/workflows/format.yml
```diff
@@ -1,21 +1,52 @@
...
+        uses: avsm/setup-ocaml@v1
+        with:
+          ocaml-version: ${{ matrix.ocaml-version }}
+
       - name: Install dependencies
-        run: opam install dune ocamlformat
+        if: steps.cache-opam.outputs.cache-hit != 'true'
+        run: opam install ocamlformat
+
+      - name: Upgrade dependencies
```
smorimoto, line 43: Same here.

.github/workflows/ci.yml
tmattio: Thanks for the review @smorimoto! I made the changes in #184

PR #186, opened 2020-09-09
Add common middlewares

tmattio:
This PR introduces new middlewares to Opium:

- `Middleware.content_length` - Add the `Content-Length` header to the response if it does not exist. This allows other middlewares (e.g. `logger` and `debugger` to display the actual request that will be sent to the client, otherwise it would be missing the header that is automatically added by the `Server_connection.run` function.

- `Middleware.method_override` - Override the HTTP method with the content of the header `_method`. This is useful when querying an endpoint from an HTML form because the `action` attribute only supports the `GET` and `POST` methods.

- `Middleware.etag` - Add the `ETag` header with the MD5 hash of the body content and returns the HTTP code 304 if the client provided an `Etag` that matches the one calculated.

It also makes the following changes to the API for consistency:

- The router middleware (previously `Router.m`) now lives with the other middlewares at `Middleware.router` for consistency.

Finally, the middlewares are now located in a `middlewares` directory. It does not impact the API or the documentation: we simply use the `(include_subdirs unqualified)` Dune's stanza to clean up the root directory.

**To Do**

- [x] Implement the ETag middleware
- [x] Complete middlewares documentation


opium_kernel/src/middlewares/middleware_method_override.ml
```diff
@@ -0,0 +1,22 @@
...
+    let content_type = Request.content_type req in
+    match req.Request.meth, content_type with
+    | `POST, Some "application/x-www-form-urlencoded" ->
+      let* method_result =
+        Request.urlencoded "_method" req
+        |> Lwt.map (fun el -> Option.map String.uppercase_ascii el)
+      in
+      let method_ =
+        match method_result with
+        | Some "PUT" -> `PUT
```
rgrinberg, line 13: Isn't there a `Method.of_string` somewhere?

rgrinberg (approved):
Looks OK apart from my nitty comment.

opium_kernel/src/middlewares/middleware_method_override.ml
tmattio: There is, but the implementation looks like this:

```
let of_string =
  function
  | "GET" -> `GET
  | "HEAD" -> `HEAD
  | "POST" -> `POST
  | "PUT" -> `PUT
  | "DELETE" -> `DELETE
  | "CONNECT" -> `CONNECT
  | "OPTIONS" -> `OPTIONS
  | "TRACE" -> `TRACE
  | s -> `Other s
```

Since the user can send any value with the `_method` field, we could end up with an invalid HTTP method. Plus, we know that the request is coming from an HTML form, so it might not make sense to allow other HTTP methods. FWIW, the implementation is on-par with https://github.com/elixir-plug/plug/blob/master/lib/plug/method_override.ex.

rgrinberg: > There is, but the implementation looks like this:

Seems like this implementation should be fixed to allow PATCH. @anuragsoni any reason why it's absent?

You can still use this implementation by the  way:

```ocaml
match Method.of_string method_result with
| (`PUT | `Delete | `Other "PATCH" as m) -> m
|  _ -> req.meth
```

Which still seems to be cleaner. At least, we have less opportunity to introduce typos in strings.

tmattio: That looks better indeed!

The Method module includes Httpaf.Method, so the implementation of the `of_string` comes from Httpaf. No idea why they don't include Patch in the variant though..

anuragsoni: > Seems like this implementation should be fixed to allow PATCH. @anuragsoni any reason why it's absent?

PATCH is absent at the moment because when I did the initial port I chose to use Httpaf.Method.standard. When I switched the router to use a Map instead of a fixed length array I moved the type to Method.t but forgot to expose PATCH in the public api. I would lean towards adding PATCH in opium's public api.

PR #187, opened 2020-09-09
Move Hmap0 to Request.Context

tmattio:
This PR is mostly an RFC.

I see two issues with the Hmap0 module at the moment:

- It is used as a top-level module and both the request and response have a record field `env` that uses it. I am not sure it makes sense that the response has an `env`, in all the frameworks I've worked with, there was only one context and I don't remember being limited by it in any way. If we remove the `env` field from the response, the Hmap0 module is only used by the request and it would be natural to find it as a submodule of `Request`.

- The name is not very instructive. It is indeed giving hints on how it works, but not how/when to use it. For a new user, I think it would be less confusing to name it something like `Context` or `Environment` (I tend to prefer the former because that's the terminology used in other frameworks).

With this in mind, I propose to move the Hmap0 module to `Request.Context`. If this makes sense for everyone, I think it would be nice to offer some helper functions in the `Request` module to manage the context more easily (i.e. getters and setters)


rgrinberg:
> I am not sure it makes sense that the response has an env, in all the frameworks I've worked with, there was only one context and I don't remember being limited by it in any way.

There is no limitation, but some things are a little more convenient to do with a response side environment. For example, if you're writing middleware that handles sending cookies. You are better off saving the cookies on the response side. I'm not too sure this convenience is worth it.

Can we see what other libraries are doing? I know Haskell's WAI indeed only has a single field. Let's take examples from well typed languages however.

At some point, I had the far fetched idea of creating a middleware based web client as well. Something like:

```ocaml
val send : client -> Request.t -> Response.t Lwt.t
```

In which case, response side `env` is required for middleware to populate extra information obtained from the response.

However, at this point, the idea would need support from other maintainers before we spend time on it.

> The name is not very instructive. It is indeed giving hints on how it works, but not how/when to use it. For a new user, I think it would be less confusing to name it something like Context or Environment (I tend to prefer the former because that's the terminology used in other frameworks).

What kind of information does the name `Context` convey? Also if you're going to rename this type, you should rename the field `env` as well. It should be `env : Env.t` or `context : Context.t`. `Hmap` can break this convention since it's just the raw data structure. In WAI, which is the other framework I'm familiar with, this field & type is also called "vault".

tmattio:
> At some point, I had the far fetched idea of creating a middleware based web client as well

That's an interesting use case! I really like the architecture of [Plug](https://github.com/elixir-plug/plug) and they use a single structure named `Connection` where they store both the request, the response, and the context. Turns out, it's really flexible and your use case would be handled nicely. But with a separated Request and Response, I guess the only way would be to keep an env on the Response side.. I'm not sure this is worth it though, this use case seems really "niche".

> What kind of information does the name Context convey?

I guess part of it is coming from conventions I've used in the past, where the "Context" is generally used to refer to the key value store shared by all the components of the processing pipeline and available once the request arrives at the "controller" (or handler). But I might be biased, I might have worked on too many GraphQL backends, could be that and the name "Context" isn't that general after all 😅 

> Can we see what other libraries are doing?

Yes! I'll do some more research.

rgrinberg:
I'd such suggest checking out Finagle from the Scala folks. The authors of Finagle wrote the paper Opium is inspired by. In this paper Requests and Responses are kept separate.

tmattio:
Will do, thanks for the reference!

anuragsoni:
> At some point, I had the far fetched idea of creating a middleware based web client as well. Something like:

Interesting you mention that. This is something that i've also experimented with. I discovered Opium after discovering the finagle paper, and the `req -> res IO.t` model felt appealing. I had a prototype of client integration with opium where services could be used for both server/client tasks. At that point i had a few filters in mind that i thought could be shared across both server/client, logging, middlewares to deal with timeouts, encoding/decoding data, middlewares that need to enhance request/response structure with additional metadata (could be headers), etc. While this could potentially be achieved with a single environment entity scoped within a request, having an environment for the responses seemed pretty nice as well.

I would eventually like to see some sort of client integration in opium as well as i'd like to have an easy to use client call api provided out of the box. Now that we are using httpaf, something based on httpaf would be preferable i think.

rgrinberg:
> I would eventually like to see some sort of client integration in opium as well as i'd like to have an easy to use client call api provided out of the box. Now that we are using httpaf, something based on httpaf would be preferable i think.

Indeed. How far is httpaf from having a usable client?

tmattio:
> How far is httpaf from having a usable client?

There's [Piaf](https://github.com/anmonteiro/piaf), which is really good, but it uses a fork of Httpaf that is not published on Opam. That being said the fork uses the same API, so it should be feasible to build a client with the original Httpaf repo.

tmattio:
> I would eventually like to see some sort of client integration in opium

As much as I'd love to have a good HTTP client in OCaml (Cohttp is good, but too low level IMO), I'm not sure it should live in Opium. I mean, it makes sense to use the same building blocks (filters and services), but I would make it a separate package. People who need an HTTP client don't necessarily want to install a server, and vice-versa.

> Can we see what other libraries are doing?

I checked Finagle yesterday. They keep a context in both the Request and Response and they call it "Context". I didn't dive deeper though, I'll keep researching.

anuragsoni:
> As much as I'd love to have a good HTTP client in OCaml (Cohttp is good, but too low level IMO), I'm not sure it should live in Opium. I mean, it makes sense to use the same building blocks (filters and services), but I would make it a separate package. People who need an HTTP client don't necessarily want to install a server, and vice-versa.

I wasn't suggesting to make it the same package. But leaving the building blocks to allow for that use-case might be something that's worth maintaining? That said i'll defer to the decision that's made in this case as I don't want to block something based on a hypothetical future development :) 

> Indeed. How far is httpaf from having a usable client?

In its current form its lacking ssl/tls support (this is true for the server as well). Its possible to work around it by wrapping httpaf with something like conduit/lwt + lwt_ssl combination, etc, but then we'd need to maintain our own driver for httpaf. I did something similar with async via https://github.com/anuragsoni/async-http when i had a need for making client calls. I'd like to contribute upstream and add support for this, just need to find time and an approach that suits their desired approach.

rgrinberg:
>  I mean, it makes sense to use the same building blocks (filters and services), but I would make it a separate package. People who need an HTTP client don't necessarily want to install a server, and vice-versa.

That's why in dune we support multiple packages out of the same repo ;). It's quite a pain to create repos for every little subproject like this.

tmattio:
> I wasn't suggesting to make it the same package. But leaving the building blocks to allow for that use-case might be something that's worth maintaining? 

In this case, completely onboard!

Incidentally, I'm working on scraping stuff at the moment, I'm working in Python, because the ecosystem is just too big to ignore, but I contemplated the idea of building an equivalent of Scrapy in OCaml, and thought that the Rock architecture would fit very nicely for a scraping framework!

I have also been thinking about the documentation of the Rock module recently: it's a nice foundation to build modular servers (and apparently clients), but I don't think most users will care about it, for them, it's just an implementation detail and not something they should be exposed to (I mean the filter/service modules, not the handler/middlewares).

So extracting the Rock architecture in a separate package would serve two purposes IMO:

- Provide the foundations for a client library with the same architecture as Opium
- Remove implementation details from Opium's API that users don't necessarily care about. If they do care about it, they could use the lower-level library containing 

Is this in line with what you envisioned?

I'm just not certain what features such a library could provide though. The Request and Response modules are specific to servers, and so are the Handler and Middleware modules. So the only extractable part at the moment would be:

```
module Service = struct
  type ('req, 'res) t = 'req -> 'res Lwt.t

  let id req = Lwt.return req
end

module Filter = struct
  type ('req, 'rep, 'req_, 'rep_) t = ('req, 'rep) Service.t -> ('req_, 'rep_) Service.t
  type ('req, 'rep) simple = ('req, 'rep, 'req, 'rep) t

  let ( >>> ) f1 f2 s = s |> f1 |> f2
  let apply_all filters service = ListLabels.fold_left filters ~init:service ~f:( |> )
end
```

Which I'm not sure is very useful as a standalone library. What more could this library do?

tmattio:
Going to close this for now. The discussion this PR triggered lead to #204 and I renamed `Hmap0` to `Context` in there. I think the more controversial aspect was the removal of `env` from Response, which I haven't done in #204, so we can come back to in the future if needed.

PR #188, opened 2020-09-11
Support multipart form data

tmattio:
This adds support to decode `multipart/form-data`-encoded requests.

The parsing is delegated to https://github.com/cryptosense/multipart-form-data, so nothing too fancy here.

I added an example of a file upload using the provided functions. The example comes in two variants:

- `examples/file_upload/main.exe` that provides a nice UX and uses TailwindCSS and AlpineJS
image

- `examples/file_upload/simple.exe` that is much simpler and only uses what's necessary to implement the file upload
image

It would be nice to have support for file uploads directly in Opium, but I think it should live in Opium_kernel and we don't have a solution to implement IO functions at the moment. At the same time, passing an IO module doesn't make for a very nice API, so in the meantime, people can just copy-paste the example.


rgrinberg (approved):
Looks good. However, Lwt_stream is gross and we should avoid/discourage its use. Is it possible to use multipart without it?

tmattio:
I initially wanted to use https://github.com/dinosaure/multipart_form, which does not use `Lwt_stream`, but it's not released on Opam. I'd be happy to use it instead of `multipart-form-data` if @dinosaure releases it on Opam at some point though.

But I'm curious: what's wrong with Lwt_stream that we should avoid using it?

dinosaure:
@tmattio I can do a release of it if you want 👍

tmattio:
@dinosaure That would be awesome! I'll update the code to use your implementation when it's on Opam :)

rgrinberg:
> what's wrong with Lwt_stream that we should avoid using it?

[It has poor semantics](https://github.com/ocsigen/lwt/issues/250#issuecomment-222146474)

PR #189, opened 2020-09-11
Add a new package opium-testing

tmattio:
This PR introduces a new package `opium-testing` that provides helpers to test Opium applications.

It's a very lightweight package, with only a dozen testing functions, but with the helper functions introduced in #179, it makes testing Opium apps straightforward. Here's a test example from the documentation:

```ocaml
let test_can_create_user () =
  let req =
    Request.of_json ~body:(`Assoc [ "name", `String "Tom" ]) "/create-user" `POST
  in
  let+ res = handle_request req in
  check_status `Found res.status
;;
```


opium-testing/src/opium_testing.ml
```diff
@@ -0,0 +1,114 @@
+module Testable = struct
+  let status = Alcotest.of_pp Opium_kernel.Status.pp_hum
```
tmattio, line 2: Should be `pp` not `pp_hum`

opium-testing/src/opium_testing.ml
```diff
@@ -0,0 +1,114 @@
...
+  let meth = Alcotest.of_pp Opium_kernel.Method.pp_hum
+  let version = Alcotest.of_pp Opium_kernel.Version.pp_hum
+  let body = Alcotest.of_pp Opium_kernel.Body.pp_hum
```
tmattio, line 5: The body won't be tested here if it's a stream. Maybe that's not a problem if the `check_body` function does the job properly.

opium-testing/src/opium_testing.ml
```diff
@@ -0,0 +1,114 @@
...
+  let request = Alcotest.of_pp Opium_kernel.Request.pp_hum
+  let response = Alcotest.of_pp Opium_kernel.Response.pp_hum
+end
+
+let handle_request app =
```
tmattio, line 10: Maybe be worth it to split up `Server_connection.run` instead of duplicating the logic here.

opium-testing/src/opium_testing.mli
```diff
@@ -0,0 +1,149 @@
...
+(** [check_response' ?msg t1 t2] checks that the response [t1] and [t2] are equal.
+
+    This is a labeled variant of {!check_response} *)
+val check_response'
+  :  ?msg:string
+  -> expected:Opium_kernel.Response.t
+  -> actual:Opium_kernel.Response.t
+  -> unit
+
+(** {3 [assert_body_contains]} *)
```
tmattio, line 146: Change to `check_body_contains`

rgrinberg (approved):
Looks good. Although you should really look into `ppx_expect` for unit testing. It's much easier to use.

tmattio:
> Although you should really look into ppx_expect for unit testing. It's much easier to use.

I've been using it to test some JSOO apps, and indeed there's less boilerplate. Although to be honest, I'm not a fan of the all the PPX magic, but it's just a matter of test I guess.

PR #190, opened 2020-09-11
Add support for cookies

tmattio:
This PR brings back support for cookies.

Since Httpaf does not support cookies as-is, we have to implement our own cookie management. The `Cookie` module in this PR is based on `ocaml-cookie`, a backend-agnostic implementation of Cookies. I didn't use it as a dependency for two reasons:

- I added support for signed cookies
- I changed the module's API to be more cohesive with the rest of Opium's API

Also, the implementation is small enough that it can fit in one file, so it seemed like an overhead to add it as a dependency.

@ulrikstrid tagging you here FYI and to thank you for your work, `ocaml-cookie` is a really nice package!
I added `ocaml-cookie`'s license on top of the files, but let me know if there's anything else you'd like me to do, or if you have any concern 🙂 

**To Do**

- [x] Finish the documentation
- [x] Add some tests for the new Request / Response functions


ulrikstrid:
I would be very open to a PR changing the API in ocaml-cookie. I don't think there are many users of it yet so it should be fine.
I created it because I wanted a less ad-hoc way of dealing with cookies when writing httpaf based servers.

rgrinberg (approved):
Looks good to me. I'd prefer to work with upstream if they're flexible (as Ulrik is here).

ulrikstrid:
I would love for there to be one nice cookie implementation across all frameworks so that we can consolidate the effort

tmattio:
> I would be very open to a PR changing the API in ocaml-cookie.

That's great! Would you be open to a PR that adds support for signed cookies as well?

I'll prepare a PR to `ocaml-cookie`, but if that's ok with the both of you, I'll do it in parallel as to not have it as a blocker for other things I'm working on (middlewares and sessions). If it happens that this PR is ready to merge before the changes have been upstreamed to `ocaml-cookie`, I'd prefer to make another PR to synchronize Opium.

rgrinberg:
Sounds fine to me. Let’s just make sure our next release uses ocaml-cookie (if it’s ready by then)

tmattio:
I think this is ready to merge, so I'll make the changes to use `ocaml-cookie` once we've upstreamed.

@ulrikstrid I won't be able to make a PR next week, but probably the week after. Let me know if you're ok with me including the cookie signing part 🙂

ulrikstrid:
Please add signed cookies as well, it looks very useful. Especially since my main usecase is auth 😁

tmattio:
Awesome, will do!

PR #204, opened 2020-10-01
Refactor opium_kernel to rock

tmattio:
This PR refactors `opium_kernel` to `rock` and moves every high-level feature to opium.

The rock package is now only composed of primitives that allow sharing handlers and middlewares. It comes with very few dependencies and offers what I think is a good re-usable foundation for other web frameworks that follow the same architecture as Opium (e.g. re-web, morph, sihl) and will allow sharing middlewares and handlers between each of these frameworks.

It's also now a good candidate to build an HTTP client: all the server-specific features and dependencies now live in opium. (except the Server_connection, but we could imagine providing a similar module for HTTP clients).

As `rock` is a lower level package whose users are going to be "expert users", I removed the HTTP primitives (Headers, Status, Version, etc.). Users can use Httpaf instead. I moved the primitives to Opium though, as we don't want end-users to fallback to Httpaf's documentation.

I sneaked in the renaming of Hmap0 to Context. I let the context in [Response.t] however, as I think it was the controversial point.

**Todo**

- [X] Implement back the Unix static middleware
- [X] Inline documentation for Headers, Status, Method and Version
- [x] Add back the Rock.App module
- [x] Fix examples
- [x] Update packages descriptions
- [x] Complete missing documentation


tmattio:
@rgrinberg @anuragsoni I think this is ready to review now.

I didn't change much code (apart from the examples), I mostly moved thing around, so I think there's not a lot of value to go through the huge diff, especially if you're short on time.

Instead, here's a list of things that changed that we might want to discuss.

## Moved features from `Opium_kernel` to `Opium`

To reduce the number of dependencies in `rock`, I moved a lot of features from `Opium_kernel` to `Opium`. I first tried to remove only the features that were relying on external dependencies, but the resulting API seemed weird and very incomplete.

So I decided to keep only the core of `Rock`. The resulting API is mostly composed of type definitions and a `Server_connection` module that remains untouched. Even if the surface API now extremely small, I found that `rock` was still a good base to build HTTP servers (see https://github.com/rgrinberg/opium/blob/bf40fe7ce75934970d63d7de76590239f7258dec/example/rock_server/main.ml),
while having just a few dependencies.

Most notable, I moved the following features/modules to `Opium_kernel`:

- `Route` and `Router`
- All the middlewares
- Request / Response helper functions

## Renaming Hmap0 to Context

As I mentioned in the PR description, I renamed Hmap0 to Context. The term context is found in multiple other web frameworks and I think new users will appreciate that we use existing terminology.

I didn't touch the `env` field from the `Request.t` and `Response.t` however.

In `rock`, `Context` exposes all the function of the `Hmap.Make` functor. In `opium`, I removed some of them (the ones I thought weren't useful) to slim the API a bit.

## Remove Opium.Std

With the move of all the features from `Opium_kernel` to `Opium`, I revisited the top-level API of `Opium`: I removed `Opium.Std` so there is a single API.
The app builder functions still exist in the `App` module, but they are not exposed at the top level anymore.

## Inline documentation of re-exposed modules

Lots of modules in `Opium` are supersets of other libraries' modules. For instance:

- `Version` is a superset of `Httpaf.Version`
- `Method` is a superset of `Httpaf.Method`
- `Headers` is a superset of `Httpaf.Headers`

Others are including modules, keep the same types, but rework the API, for instance:

- `Context` reworks the API of `Rock.Context`

When navigating the documentation, I found it very unfriendly for newcomers to have to jump to another library's documentation to read the function definition of functions that we are exposing in our own API.
The solution I came up with is to essentially copy the documentation of the modules we re-expose. This improves the documentation drastically but is not very satisfying, I wonder if there would be a better alternative.

rgrinberg (approved):
Looks good. CI is still failing though

tmattio:
I'll fix that, and I still have some documentation to write before this is 100% done, but was waiting for feedback before I complete it 🙂 

Do you have any concerns about the copy-pasting of documentations (i.e. "Inline documentation of re-exposed modules
")? This is actually the only point where I'm not really sure this goes in the right direction. I figure this is a common problem, so I was wondering if you happen to know an elegant solution for it.

rgrinberg:
> Do you have any concerns about the copy-pasting of documentations (i.e. "Inline documentation of re-exposed modules
")? This is actually the only point where I'm not really sure this goes in the right direction. I figure this is a common problem, so I was wondering if you happen to know an elegant solution for it.

I don't see anything unfriendly about linking to the docs of another package. One of the reasons why odoc was introduced was to allow that. I think it would make sense for us to upstream some of our extensions to these modules however.

tmattio:
Odoc doesn't link to the doc of other packages though, at least not when I build the doc locally, but I might be doing something wrong?

Most of the modules I'm referring to are Httpaf modules and looking at the repository's issues and PRs, I have little faith that we can upstream the changes 😕

anuragsoni:
> Odoc doesn't link to the doc of other packages though, at least not when I build the doc locally, but I might be doing something wrong?

if you reference a module from Httpaf (just an example, could be any other module from your dependencies), odoc should create a hyperlink that allows you to jump to the definition. For that to work you'd need to either generate docs in a duniverse sandbox so all the dependencies are actually present and cross-references, or use `odig` and generate the documentation set for opium and its dependencies. That will get you a landing page with an index of all dependencies used by opium, and fully cross-referenced odoc output that can be nice for the users.

tmattio:
@anuragsoni Good to know, thanks for the tips! We probably don't need to duplicate the documentation in this case, I'll get back to it and make the appropriate changes.

rgrinberg:
> Most of the modules I'm referring to are Httpaf modules and looking at the repository's issues and PRs, I have little faith that we can upstream the changes

Give it a try. If the changes are reasonable, I don't think we'll have many problems getting them in.

rgrinberg:
@tmattio how soon are we from merging? I'd like to do some work on top of this.

tmattio:
@rgrinberg Sorry for the delay, things have been busy lately, and I didn't get much time to get back to it.

I've been using this branch in my project for a month now, so I'm confident this is stable enough. The only missing piece is the documentation of a few functions.

If you don't mind me merging it with the duplicated `mli` files that I will revisit later, I can make the remaining changes tomorrow.

rgrinberg:
I don't mind to merge now. Thanks for all your work on this

tmattio:
Sure, feel free to merge now if this is blocking you, I can make another PR for the doc (and the CI failure) 🙂

tmattio:
Alright, merging now! Will make a new PR for the duplicated docstrings soon.

PR #220, opened 2020-11-16
Simplification/Deforestation

rgrinberg:
General simplifications

### DIFF
--- a/opium/src/import.ml
+++ b/opium/src/import.ml
@@ -16,6 +16,16 @@ module List = struct
       | Some _ as result -> result
       | None -> find_map ~f l)
   ;;
+
+  let replace_or_add_to_list ~f to_add l =
+    let rec aux acc l found =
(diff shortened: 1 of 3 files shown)


tmattio (approved):

opium/src/import.ml
```diff
@@ -16,6 +16,28 @@ module List = struct
       | Some _ as result -> result
       | None -> find_map ~f l)
   ;;
+
+  let replace_or_add_to_list ~f to_add l =
```
rizo, line 20: Maybe this could be just `List.replace_or_add`? The `_to_list` suffix is now redundant.

rizo (commented):
Looks good functionally! 👍 

I wonder if, in some places, using `Seq` (or other iterator models) would be a better choice long-term to achieve automatic fusion and help with code clarity.

PR #227, opened 2020-11-27
Fix dependencies

tmattio:


rgrinberg (approved):

tmattio:
The cache for the `opam` is corrupted (wrong versions of the OCaml compiler), so I disable it for now. Github invalidates the cache after 7 days, so I'll make a PR to re-enable it then.

PR #232, opened 2020-12-03
Add odoc with-doc dependency

reynir:
Should enable us to `opam install --with-doc opium`.

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -30,7 +30,8 @@
   hmap
   httpaf
   lwt
-  sexplib0))
+  sexplib0
+  (odoc :with-doc)))
 
 (package
  (name opium)
@@ -59,6 +60,7 @@
   re
   uri
   multipart-form-data
+  (odoc :with-doc)
   (alcotest :with-test)
   (alcotest-lwt :with-test)))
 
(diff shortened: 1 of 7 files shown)


tmattio:
Thanks @reynir!

The dependencies are listed in `dune-project` and the `opam` files are generated by Dune. In addition, we use lock files for the CI, so we need to regenerate them when changing the dependencies (with `opam lock -y .`)

You'll probably want to add `odoc` to `rock` and `opium-testing` as well.

reynir:
I added odoc {with-doc} in the proper place and for all packages.

`opam lock` does not seem to understand the `-y` flag, and without the diff is rather large. I did not commit that since I'm not sure it's right.

tmattio:
I generated the lock files (the lock files you generated did not include test dependencies)

tmattio:
Thank you @reynir 🎉

reynir:
Do you know why they didn't? Anyway, thank you!

tmattio:
I think they don't get generated if you didn't install them in you current switch (so you need to use `opam install` with `--with-test`)

PR #234, opened 2020-12-03
Add example/json_response/

### LINKED ISSUE #233: example/json_response/ referenced but does not exist
In example/README.md

reynir:
It returns a json object with fields os-type and ocaml-version with values from `Sys`.

Fixes #233

### DIFF
--- a/example/json_response/README.md
+++ b/example/json_response/README.md
@@ -1,0 +1,14 @@
+# JSON Response Example
+
+```
+dune exec example/json_response/main.exe
+```
+
+This is an example of a JSON response.
+
+The server offers an endpoint `/` that serves a single JSON object.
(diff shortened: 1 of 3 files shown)


tmattio:
Thanks @reynir!

Could you add a README that describes the example?

The examples are probably the first entry point for beginners, so the README could probably contain:
- A description of how the example work (the use of `Response.of_json`)
- A link to the documentation of Yojson

The existing examples might not be as descriptive, but they should definitely be

tmattio:
Here's an example of a README: https://github.com/rgrinberg/opium/tree/master/example/file_upload

tmattio (approved):

tmattio:
Looks good, thanks a lot! Will just wait for the CI to pass.

tmattio:
Thanks! 🎉

PR #236, opened 2020-12-03
static_serve: comment does not reflect the code

reynir:
This PR removes the comment in `example/static_serve/main.ml` which does not reflect what the program actually does.

I think the mentioned example is worth having, though.


tmattio (approved):
The bug mentioned will probably be fixed by #215, so it's fine to just remove it.

PR #237, opened 2020-12-03
example/user_auth: implement app

reynir:
The `user_auth` program was a noop. This implements a page that welcomes logged in users and otherwise tell them they are unauthorized.

I am not too familiar with basic authentication. I feel `index_handler` is probably the wrong place to add the `www-authenticate` header.

### DIFF
--- a/example/user_auth/main.ml
+++ b/example/user_auth/main.ml
@@ -94,3 +94,23 @@ let m auth =
 ;;
 
 let user { Request.env; _ } = Opium.Context.find Env.key env
+
+let index_handler request =
+  let challenge = `Basic "something" in
+  match user request with
+  | None ->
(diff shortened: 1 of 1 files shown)


reynir:
I figured out how to add the header in the middleware (I learned something! yay). It checks the resulting response if it has status code `` `Unauthorized` and sets the header in that case.

tmattio:
Thanks a lot for doing this @reynir!

Opium ought to provide basic functions for authentication, so this PR motivated me to work on this. I've extracted the code of your revisited example that could live in Opium and kept the rest. I created a new PR: https://github.com/rgrinberg/opium/pull/238, so I'll close this one.

Thanks again 🙂

PR #238, opened 2020-12-03
Basic auth middleware

tmattio:
This PR adds a `basic_auth` middleware to protect handlers with an authentication mechanism.

It also adds an `Auth` module with functions to encode/decode `Authorization` header values.

The `basic_auth` middleware has type

```ocaml
val basic_auth
  :  ?unauthorized_handler:Rock.Handler.t
  -> key:'a Context.key
  -> realm:string
  -> auth_callback:(username:string -> password:string -> 'a option)
  -> unit
  -> Rock.Middleware.t
```

I believe it is flexible enough and will allow users to build their own authentication logic, without having to deal with the boilerplate logic.

**TODO**

- [x] Document the `Auth` module
- [ ] Support `Bearer`, and maybe cookie?

### DIFF
--- a/example/user_auth/README.md
+++ b/example/user_auth/README.md
@@ -9,16 +9,13 @@ This example implements a very simple authentication system using `Basic` authen
 The middleware stores the authenticated user in the request's context, which can be retrieved in the handlers.
 
 The username and password for the authentication are `admin` and `admin`.
-The corresponding `Basic` header is `Basic YWRtaW46YWRtaW4=`.
 
-You can test that you are unauthorized to access the `/` endpoint without the correct authorization header:
-
-```sh
-curl http://localhost:3000/ -X GET
-```
-
-And that you are allows to access when you provide it:
-
+You can test that you are authorized to access the `/` endpoint with the correct `Authorization` header:
 ```sh
 curl http://localhost:3000/ -X GET --user admin:admin
+```
+
+And that you are not allows to access it when you don't provide the a valid `Authorization` header:
+```sh
+curl http://localhost:3000/ -X GET
 ```
--- a/example/user_auth/README.md
+++ b/example/user_auth/README.md
@@ -20,5 +20,5 @@ curl http://localhost:3000/ -X GET
 And that you are allows to access when you provide it:
 
 ```sh
-curl http://localhost:3000/ -X GET -H "Authorization: Basic YWRtaW46YWRtaW4="
+curl http://localhost:3000/ -X GET --user admin:admin
 ```
--- a/example/user_auth/README.md
+++ b/example/user_auth/README.md
@@ -7,3 +7,18 @@ dune exec example/user_auth/main.exe
 This example implements a very simple authentication system using `Basic` authentication.
 
 The middleware stores the authenticated user in the request's context, which can be retrieved in the handlers.
+
+The username and password for the authentication are `admin` and `admin`.
+The corresponding `Basic` header is `Basic YWRtaW46YWRtaW4=`.
+
+You can test that you are unauthorized to access the `/` endpoint without the correct authorization header:
+
+```sh
+curl http://localhost:3000/ -X GET
+```
+
+And that you are allows to access when you provide it:
+
+```sh
+curl http://localhost:3000/ -X GET -H "Authorization: Basic YWRtaW46YWRtaW4="
+```
--- a/example/user_auth/main.ml
+++ b/example/user_auth/main.ml
@@ -1,96 +1,55 @@
 open Opium
-open Sexplib0.Sexp_conv
 
-module Auth = struct
-  (* https://github.com/mirage/ocaml-cohttp/blob/35e1386dcca759bcc955c59c7e91260f765f253b/cohttp/src/auth.ml *)
-  (*{{{ Copyright (c) 2012 Anil Madhavapeddy <anil@recoil.org>
-   *
-   * Permission to use, copy, modify, and distribute this software for any
-   * purpose with or without fee is hereby granted, provided that the above
-   * copyright notice and this permission notice appear in all copies.
-   *
-   * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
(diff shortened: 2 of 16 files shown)


tmattio:
@jerben I'm curious, do you have any authentication layer in Sihl? If so, would it be compatible with the changes in this PR?

example/user_auth/main.ml
```diff
@@ -1,96 +1,55 @@
...
-type user = { username : string (* ... *) } [@@deriving sexp]
-
-(* My convention is to stick the keys inside an Env sub module. By not exposing this
-   module in the mli we are preventing the user or other middleware from meddling with our
-   values by not using our interface *)
-module Env = struct
-  (* or use type nonrec *)
-  type user' = user
+module Env_user = struct
+  type user' = User.t
```
reynir, line 20: Maybe we should just call it `user` instead of `user'` since the old `user` is now `User.t`.

example/user_auth/README.md
```diff
@@ -7,3 +7,18 @@ dune exec example/user_auth/main.exe
...
+You can test that you are unauthorized to access the `/` endpoint without the correct authorization header:
+
+```sh
+curl http://localhost:3000/ -X GET
+```
+
+And that you are allows to access when you provide it:
+
+```sh
+curl http://localhost:3000/ -X GET -H "Authorization: Basic YWRtaW46YWRtaW4="
```
reynir, line 23: You can also use `--user admin:admin` with curl.

opium/src/middlewares/middleware_basic_auth.ml
```diff
@@ -0,0 +1,30 @@
...
+        Response.of_plain_text "Forbidden access" ~status:`Unauthorized
+        |> Response.add_header ("WWW-Authenticate", Auth.string_of_challenge (Basic realm))
+        |> Lwt.return)
+  in
+  let filter handler ({ Request.env; _ } as req) =
+    let open Lwt.Syntax in
+    let+ resp =
+      match Request.authorization req with
+      | None -> unauthorized_handler req
+      | Some (Other _) -> handler req
```
reynir, line 13: Should we handle other authentication methods this way? We have effectively not authenticated the user but still continue. The following curl snippet will avoid the `unauthorized_handler`, but in the example it a `Not_found` exception is raised because the user key is not set.

```
curl localhost:3000 -H "Authorization: Basic T0NhbWwK" # base64 for "OCaml" - no ':'
```

I think `unauthorized_handler` should be called setting bad request status or so.

opium/src/request.mli
```diff
@@ -424,6 +424,17 @@ val content_type : t -> string option
     [Content-Type] set to [content_type]. *)
 val set_content_type : string -> t -> t
 
+(** {3 [authorization]} *)
+
+(** [authorization t] returns the value of the header [Authorization] of the request [t]. *)
+val authorization : t -> Auth.Credential.t option
+
+(** {3 [set_authorization]} *)
```
reynir, line 432: This is only relevant when making requests I think. I'm not very familiar with opium, but is it something that we want to expose or will be used?

reynir (changes requested):
I look over the changes and have some comments about the example, but also how 'other' authentication methods are handled.

jerben:
> @jerben I'm curious, do you have any authentication layer in Sihl? If so, would it be compatible with the changes in this PR?

@tmattio At the moment, we have that in the projects themselves. Your approach looks good and we could definitely use the middleware as proposed, I especially like the configurable rock handler :+1:

opium/src/opium.mli
```diff
@@ -221,4 +222,22 @@ module Middleware : sig
...
+      The requests have to provide an [Authorization] header with the format
+      [Basic = <credentials>]. [auth_callback] is called with the username and password
+      extracted from the credentials. If the user does not contain a valid [Authorization]
+      header, or if the [auth_callback] returns [None], the request is redirected to
+      [unauthorized_handler] (by default, returns a "Forbidden access" message). *)
+  val basic_auth
+    :  ?unauthorized_handler:Rock.Handler.t
+    -> key:'a Context.key
+    -> realm:string
+    -> auth_callback:(username:string -> password:string -> 'a option)
```
tmattio, line 240: This should be `(username:string -> password:string -> 'a option Lwt.t)`

opium/src/request.mli
tmattio: This is especially useful to build requests for unit tests

opium/src/middlewares/middleware_basic_auth.ml
tmattio: Do you mean handling everything with `handler req`, or handling other auth methods with `unauthorized_handler req`?

reynir: Uh, I'm a bit confused by the question - but I'm happy with the change in the latest commit, so I consider this resolved.

tmattio: Ok great, I was confused by the opening question as well, but I see we had the same design in mind 🙂 
`Some (Other _) -> handler req` was an oversight

example/user_auth/README.md
```diff
@@ -7,3 +7,18 @@ dune exec example/user_auth/main.exe
 This example implements a very simple authentication system using `Basic` authentication.
 
 The middleware stores the authenticated user in the request's context, which can be retrieved in the handlers.
+
+The username and password for the authentication are `admin` and `admin`.
+The corresponding `Basic` header is `Basic YWRtaW46YWRtaW4=`.
```
reynir, line 12: A bit of a nitpick, but the header is `Authorization`. Since we use `curl --user` now maybe this line is unnecessary?

tmattio: It is, thanks!

reynir:
It looks good to me now. Thanks for implementing it! :tada:

tmattio:
> It looks good to me now. Thanks for implementing it!

Thanks for the reviews, appreciate it! I'll investigate further to support Bearer / other auth methods before I merge it: I'm not confident enough that the current implementation generalizes well 🙂

PR #244, opened 2020-12-07
Provide of_file API to stream files as response

jerben:
Hey all,

I just finished updating Sihl to use Opium 0.19.0. This is the only API that I miss so far. I tried to find a sensible location for the read function, just let me know if you have any suggestions.

### DIFF
--- a/opium/src/body.ml
+++ b/opium/src/body.ml
@@ -1,1 +1,50 @@
 include Rock.Body
+
+let log_src = Logs.Src.create "opium.body.of_file"
+
+module Log = (val Logs.src_log log_src : Logs.LOG)
+
+exception Isnt_a_file
+
+let of_file fname =
+  let open Lwt.Syntax in
+  (* TODO: allow buffer size to be configurable *)
+  let bufsize = 4096 in
+  Lwt.catch
+    (fun () ->
+      let* s = Lwt_unix.stat fname in
+      let* () =
+        if Unix.(s.st_kind <> S_REG) then Lwt.fail Isnt_a_file else Lwt.return_unit
+      in
+      let* ic =
+        Lwt_io.open_file
+          ~buffer:(Lwt_bytes.create bufsize)
+          ~flags:[ O_RDONLY ]
+          ~mode:Lwt_io.input
+          fname
+      in
+      let+ size = Lwt_io.length ic in
+      let stream =
+        Lwt_stream.from (fun () ->
+            Lwt.catch
+              (fun () ->
+                let+ b = Lwt_io.read ~count:bufsize ic in
+                match b with
+                | "" -> None
+                | buf -> Some buf)
+              (fun exn ->
+                Log.warn (fun m ->
+                    m "Error while reading file %s. %s" fname (Printexc.to_string exn));
+                Lwt.return_none))
+      in
+      Lwt.on_success (Lwt_stream.closed stream) (fun () ->
+          Lwt.async (fun () -> Lwt_io.close ic));
+      Some (of_stream ~length:size stream))
+    (fun e ->
+      match e with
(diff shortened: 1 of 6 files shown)


opium/src/body.ml
```diff
@@ -1 +1,51 @@
 include Rock.Body
+
+let log_src = Logs.Src.create "opium.body.of_file"
+
+module Log = (val Logs.src_log log_src : Logs.LOG)
+
+exception Isnt_a_file
+
+let of_file ?(local_path = "") fname =
```
tmattio, line 9: I would remove `local_path` entirely, I don't think it makes sense to keep it if the function is exposed in the API

opium/src/body.ml
```diff
@@ -1 +1,51 @@
...
+                Log.warn (fun m ->
+                    m "Error while reading file %s. %s" fname (Printexc.to_string exn));
+                Lwt.return_none))
+      in
+      Lwt.on_success (Lwt_stream.closed stream) (fun () ->
+          Lwt.async (fun () -> Lwt_io.close ic));
+      Ok (of_stream ~length:size stream))
+    (fun e ->
+      match e with
+      | Isnt_a_file | Unix.Unix_error (Unix.ENOENT, _, _) -> Lwt.return (Error `Not_found)
```
tmattio, line 46: The exception handling should probably go in `Response`, or at least this should not be converted to a 
`Not_found`. It was only the case because we used the output of this function as the status of the response in the middleware.

opium/src/response.ml
```diff
@@ -161,6 +161,23 @@ let of_json ?version ?status ?reason ?headers ?env body =
     (body |> Yojson.Safe.to_string)
 ;;
 
+let of_file ?version ?reason ?headers ?env fname =
```
tmattio, line 164: This could take an optional `mime` argument that would override the detected one.

tmattio (commented):
Looks good, but the error handling in `Body.of_file` should be changed to not returns HTTP status, or removed completely.

I'm also wondering how this plays with `Handler.serve`, which is pretty much the same thing, except that it takes a custom read function:

- Should `of_file` be exposed in `Handler`?
- Should `Handler.serve` be exposed in `Response`, perhaps with an optional argument in `of_file`?

Also, maybe out of scope here, but it'd be nice to have the counterpart `Request.to_file` to save the content of the request as a file on the filesystem. (There's a file upload example that does that if I recall correctly).

opium/src/body.ml
jerben: Fixed

opium/src/body.ml
jerben: I agree, I return `Body.t option Lwt.t` instead

opium/src/response.ml
jerben: I was thinking that this could be provided as header but it probably is a quite common use case, fixed.

tmattio (approved):
Looks good! Just missing a changelog entry (don't know why the CI passed 🤷)

jerben:
> Looks good, but the error handling in `Body.of_file` should be changed to not returns HTTP status, or removed completely.

Done.

> I'm also wondering how this plays with `Handler.serve`, which is pretty much the same thing, except that it takes a custom read function:
> 
> * Should `of_file` be exposed in `Handler`?
> * Should `Handler.serve` be exposed in `Response`, perhaps with an optional argument in `of_file`?

Yes they indeed look similar. The goal of `Handler` is to provide a set of ready-made handlers that users can plug in?
I could allow for a optional `read` function in `Response.of_file` and use it to implement the handler.

> Also, maybe out of scope here, but it'd be nice to have the counterpart `Request.to_file` to save the content of the request as a file on the filesystem. (There's a file upload example that does that if I recall correctly).

Wold it make sense to have both ways on both `Request` and `Response`?

PR #246, opened 2020-12-08
Fix cookie header parsing by trimming cookie key

jerben:
I am not sure how many spaces we should allow between the key value pairs in the `Cookie` header, but this change fixes handling of multiple cookies with how the most common browser do it I think.

### DIFF
--- a/opium/test/cookie.ml
+++ b/opium/test/cookie.ml
@@ -1,0 +1,32 @@
+let headers = [ "Cookie", "yummy_cookie=choco; tasty_cookie=strawberry" ]
+
+let parse_cookies_of_headers () =
+  let c1, c2 =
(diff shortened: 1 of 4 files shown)


tmattio (approved):
LGTM, thanks for the test 🎉

opium/test/cookie.ml
```diff
@@ -0,0 +1,32 @@
+let headers = [ "Cookie", "yummy_cookie=choco; tasty_cookie=strawberry" ]
```
tmattio, line 1: This is not going to be run I think, you need to add the module to the dune file in `opium/test`

jerben: Thanks, fixed.

PR #250, opened 2020-12-16
Handle ECONNABORTED in accept

reynir:
I had an opium server crash with an error `Fatal error: exception Unix.Unix_error(Unix.ECONNABORTED, "accept", "")`. This PR handles this error case.

### DIFF
--- a/opium/src/app.ml
+++ b/opium/src/app.ml
@@ -50,9 +50,14 @@ let run_unix_multicore ?middlewares ~host ~port ~jobs handler =
     (let+ () = Lwt_unix.bind socket listen_address in
      Lwt_unix.listen socket (Lwt_unix.somaxconn () [@ocaml.warning "-3"]));
   let rec accept_loop socket instance =
-    let* socket', sockaddr' = Lwt_unix.accept socket in
-    Lwt.async (fun () -> connection_handler sockaddr' socket');
-    accept_loop socket instance
+    Lwt.try_bind
+      (fun () -> Lwt_unix.accept socket)
+      (fun (socket', sockaddr') ->
+        Lwt.async (fun () -> connection_handler sockaddr' socket');
+        accept_loop socket instance)
+      (function
+        | Unix.Unix_error (Unix.ECONNABORTED, _, _) -> accept_loop socket instance
+        | e -> Lwt.fail e)
   in
   for i = 1 to jobs do
     flush_all ();
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+# Unreleased
+
+## Fixed
+
+- Handle ECONNABORTED in accept (#250)
+
 # 0.20.0
 
 ## Added


opium/src/app.ml
```diff
@@ -50,9 +50,14 @@ let run_unix_multicore ?middlewares ~host ~port ~jobs handler =
...
-    let* socket', sockaddr' = Lwt_unix.accept socket in
-    Lwt.async (fun () -> connection_handler sockaddr' socket');
-    accept_loop socket instance
+    Lwt.try_bind
+      (fun () -> Lwt_unix.accept socket)
+      (fun (socket', sockaddr') ->
+        Lwt.async (fun () -> connection_handler sockaddr' socket');
+        accept_loop socket instance)
+      (function
+        | Unix.Unix_error (Unix.ECONNABORTED, _, _) -> accept_loop socket instance
```
tmattio, line 59: We probably want to handle these as well:

```ocaml
      | Unix.Unix_error (Unix.EBADF, _, _)
      | Unix.Unix_error (Unix.ENOTCONN, _, _)
      | Unix.Unix_error (Unix.ECONNREFUSED, _, _)
      | Unix.Unix_error (Unix.ECONNRESET, _, _)
```

I'm curious how Lwt treats errors in `establish_server_with_client_socket` also. It'd be good to have the same behavior

reynir: It seems to me `Lwt.establish_server_with_client_socket` doesn't handle `ECONNABORTED`?! https://github.com/ocsigen/lwt/blob/master/src/unix/lwt_io.ml#L1599

I think we want to fail on `EBADF` still. I don't find `ENOTCONN`, `ECONNREFUSED` or `ECONNRESET` in the list of errors in accept(2). We should probably handle `ECONNRESET` at least where it might raise. I will look into that. Is `ECONNREFUSED` not only relevant for clients?

tmattio: > I don't find ENOTCONN, ECONNREFUSED or ECONNRESET

Indeed, we don't need to care about those then, let's only consider `ECONNABORTED` 👍 

> It seems to me Lwt.establish_server_with_client_socket doesn't handle ECONNABORTED?!

That's surprising.. Can you reproduce the crash consistently? If so, does it reproduce with the single-core version?

reynir: No, I don't have a clue how to reproduce it. I exposed the opium server running on a FreeBSD machine to the internet. After 1-2 weeks I had the crash. It was running single-core.

I reported the same for establish_server upstream: https://github.com/ocsigen/lwt/issues/829

tmattio: Ah! Let's wait and see what Lwt folks say about it then, we'll implement the same fix

reynir: Actually, I am able to reproduce somewhat consistently on the FreeBSD box by running `nmap -p 3000` on my laptop targeting the FreeBSD box over the internet. That crashes it more often than not, but not consistently. I cannot reproduce locally on my Linux laptop.

I'm not able to pin this branch due to the conf-libev dependency. I'll pin to an earlier version with my path..

tmattio: > I'm not able to pin this branch due to the conf-libev dependency.

`conf-libev` is not available on FreeBSD?

tmattio: Making it mandatory was probably not a good idea, I'll remove that.

anuragsoni: > I'm not able to pin this branch due to the conf-libev dependency. I'll pin to an earlier version with my path..

I'll try this on my freebsd install later, but I think `libev` is available for install on freebsd as a port https://www.freshports.org/devel/libev

I think once that's installed `conf-libev` should install successfully.

reynir: I think libev is available on FreeBSD, I just don't have permissions on that box to install it on that box :-)

anuragsoni: > I think libev is available on FreeBSD, I just don't have permissions on that box to install it on that box :-)

Makes sense. That said, I think it'd be nice to not enforce a requirement on `conf-libev` in the opium package :-) I'm ➕ 1 with @tmattio on removing it as a dependency. The lwt docs mention libev and we should allow the end-user to decide if they can/can't use libev on their platform.

reynir: Using opium 0.19.0 and the lwt patch I was not able to reproduce after 1000 tries \o/ https://github.com/ocsigen/lwt/pull/830#issuecomment-747475969

reynir:
This is several years out of date, and I don't know if it's still relevant. Please reopen or adapt if you find it useful.

PR #251, opened 2020-12-16
Fix full splat `**`

reynir:
This fixes #249 and adds tests.

The splat values are in reverse order, and I think that's a mistake. However, that's the existing behavior of single splats. I will make a separate PR for this.

PR #253, opened 2020-12-19
Do pct_decode for Fullsplat as well

reynir:
And add a test case.


reynir:
This fixes the bug I introduced in #251

anuragsoni (approved):
Looks good to me!

anuragsoni:
Thanks!

Issue #256, opened 2020-12-27
Base64 dependency should be >= 3.0.0


roddyyaga:
See title; (details of the breaking changes in 3.x are [here](https://github.com/mirage/ocaml-base64/blob/master/CHANGES.md), most obviously it was renamed from `B64`).

rgrinberg:
Fixed in #257

PR #257, opened 2020-12-28
Fixes dependency versions for base64 and result

### LINKED ISSUE #255: Issue with result dependency version
I get errors when trying to build with `result` pinned to version 1.4 (1.5 is the most recent version):

```
$ opam switch create . --empty
$ opam pin add result 1.4
$ opam install . --deps-only
$ dune build .
File "opium/src/cookie.ml", line 298, characters 7-23:
298 |     |> Result.to_option
             ^^^^^^^^^^^^^^^^
Error: Unbound value Result.to_option
```

roddyyaga:
Fixes #255 and #256

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 Opium
 =====
 
-The current master branch is a WIP port to httpaf. If you are looking for the last version published to opam (that was using Cohttp), please take a look at https://github.com/rgrinberg/opium/tree/0.18.0
+Since version 0.19.0, Opium uses httpaf. The last version that used Cohttp can be found at https://github.com/rgrinberg/opium/tree/0.18.0
 
 ## Executive Summary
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -1,7 +1,7 @@
 Opium
 =====
 
-The current master branch is a WIP port to httpaf. If you are looking for the last version published to opam (that was using Cohttp), please take a look at https://github.com/rgrinberg/opium/tree/0.18.0
+Since version 0.19.0, Opium uses httpaf. The last version that used Cohttp can be found at https://github.com/rgrinberg/opium/tree/0.18.0
```
anuragsoni, line 4: These changes should be made to `README.cpp.md` otherwise they'll get over-written when that file gets promoted to `README.md`

PR #262, opened 2021-02-20
Cookie backed sessions

jerben:
This is just a draft of cookie backed sessions based on how the Play framework does it. (https://www.playframework.com/documentation/2.8.x/JavaSessionFlash).

It needs tests and some more thought before it can be merged.

### DIFF
--- a/opium/src/middlewares/middleware_cookie_session.ml
+++ b/opium/src/middlewares/middleware_cookie_session.ml
@@ -1,0 +1,117 @@
+let log_src =
+  Logs.Src.create ~doc:"Opium middleware for cookie-based sessions" "opium.session"
+;;
+
+module Logs = (val Logs.src_log log_src : Logs.LOG)
+module Map = Map.Make (String)
+
+module Session = struct
+  type t = string Map.t
+
+  let empty = Map.empty
+
+  let of_yojson yojson =
+    let open Yojson.Safe.Util in
+    let session_list =
+      try Some (yojson |> to_assoc |> List.map (fun (k, v) -> k, to_string v)) with
+      | _ -> None
+    in
+    session_list |> Option.map List.to_seq |> Option.map Map.of_seq
+  ;;
+
+  let to_yojson session =
+    `Assoc (session |> Map.to_seq |> List.of_seq |> List.map (fun (k, v) -> k, `String v))
+  ;;
+
+  let of_json json =
+    try of_yojson (Yojson.Safe.from_string json) with
+    | _ -> None
+  ;;
+
+  let to_json session = session |> to_yojson |> Yojson.Safe.to_string
+
+  let to_sexp session =
+    session
+    |> Map.to_seq
+    |> List.of_seq
+    |> Sexplib0.Sexp_conv.(sexp_of_list (sexp_of_pair sexp_of_string sexp_of_string))
+  ;;
+end
+
+module Env = struct
+  let key : Session.t Context.key = Context.Key.create ("session", Session.to_sexp)
+end
+
+exception Session_not_found
+
+let find key req =
+  let session =
+    try Context.find_exn Env.key req.Request.env with
+    | _ ->
+      Logs.err (fun m -> m "No session found");
+      Logs.info (fun m -> m "Have you applied the session middleware?");
+      raise @@ Session_not_found
+  in
+  Map.find_opt key session
+;;
+
+let set (key, value) resp =
+  let session =
+    try Context.find_exn Env.key resp.Response.env with
+    | _ ->
+      Logs.err (fun m -> m "No session found");
+      Logs.info (fun m -> m "Have you applied the session middleware?");
+      raise Session_not_found
+  in
+  let updated_session =
+    match value with
+    | None -> Map.remove key session
+    | Some value -> Map.add key value session
+  in
+  let env = resp.Response.env in
+  let env = Context.add Env.key updated_session env in
+  { resp with env }
+;;
+
+let persist_session signed_with cookie_key resp =
+  let session = Context.find Env.key resp.Response.env in
+  match session with
+  | None -> (* No need to touch the session cookie *) resp
+  | Some session ->
+    (* The session changed, we need to persist the new session in the cookie *)
+    let cookie_value = Session.to_json session in
+    let cookie = cookie_key, cookie_value in
+    let resp = Response.add_cookie_or_replace ~sign_with:signed_with cookie resp in
+    resp
+;;
+
+let m ?(cookie_key = "_session") signed_with =
+  let open Lwt.Syntax in
+  let filter handler req =
+    let session =
+      match Request.cookie ~signed_with cookie_key req with
+      | None -> Session.empty
+      | Some cookie_value ->
+        (match Session.of_json cookie_value with
+        | None ->
+          Logs.err (fun m ->
+              m
+                "Failed to parse value found in session cookie '%s': '%s'"
+                cookie_key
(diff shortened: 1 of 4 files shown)


opium/src/middlewares/middleware_cookie_session.ml
```diff
@@ -0,0 +1,117 @@
...
+  let to_sexp session =
+    session
+    |> Map.to_seq
+    |> List.of_seq
+    |> Sexplib0.Sexp_conv.(sexp_of_list (sexp_of_pair sexp_of_string sexp_of_string))
+  ;;
+end
+
+module Env = struct
+  let key : Session.t Context.key = Context.Key.create ("session", Session.to_sexp)
```
tmattio, line 42: Since `Session.find` takes a `Request.t` and `Session.set` takes a `Request.t`, what's the advantage of having a middleware that stores the sessions in memory?

tmattio (commented):
Awesome work @jerben! 🎉 

A few questions:

- Since `Session.find` takes a `Request.t` and `Session.set` takes a `Request.t`, what's the advantage of having a middleware that stores the sessions in memory?

- Sessions should probably come with a modular backend. I initially wanted to provide the cookie backend and the in-memory backend with Opium, and offer an API to allow users to implement their own backend. Does it make sense to do this as part of this PR?

- The next step, once we have the sessions, will be to add a flash middleware. I think you've done this in Sihl already. Is the implementation in this PR taking this use case into account?

jerben:
Thank for your quick feedback @tmattio 

> Since Session.find takes a Request.t and Session.set takes a Request.t, what's the advantage of having a middleware that stores the sessions in memory?

`Session.set` takes a `Response.t`, so this is the session API:

```ocaml
val find : string -> Request.t -> string option
val set : string * string option -> Response.t -> Response.t
```

Usage could look like this in an Opium handler:

```ocaml
...
let user_id = Opium.Session.find "user_id" req in
(* Associating a new user to the current session as artificial experiment *)
let updated_user_id = ...
Opium.Response.of_plain_text "" |> Opium.Session.set ("user_id", Some updated_user_id) |> Lwt.return
```

> Sessions should probably come with a modular backend. I initially wanted to provide the cookie backend and the in-memory backend with Opium, and offer an API to allow users to implement their own backend. Does it make sense to do this as part of this PR?

Good point. For Sihl I think we will have the cookie based session management enabled by defaults since it's stateless. So far we have been using MariaDB and Postgres backends as default. Play suggests to take care of session data that is big (> 4 KB) or sensitive yourself. 

With this PR we could easily add other backends as separate middlewares like `middlewarememory_session` and the session API `Opium.Session` stays the same. This would be a non-breaking change.

> The next step, once we have the sessions, will be to add a flash middleware. I think you've done this in Sihl already. Is the implementation in this PR taking this use case into account?

Yes that is planned :) We already have a session based middleware in Sihl, but my plan was to submit a cookie based flash message implementation for Opium and switch to that. The flash message cookie doesn't have to be signed. 
This means an app with both middlewares enabled uses two different cookies and there is no dependencies between the middlewares.
We could still implement a session based flash message middleware next to each other if the payload size limit or performance impact is an issue, but use this as default because it is stateless and has no dependencies. (https://www.playframework.com/documentation/2.8.x/JavaSessionFlash#Flash-scope)

Also we would like to contribute the CSRF and body parsing middlewares.

tmattio:
> `Session.set` takes a `Response.t`, so this is the session API:

That's what I meant, sorry. That's why I'm not sure to understand the need for the middleware - couldn't `Session.set` and `Session.get` read/write the cookie from the request/response directly?

> With this PR we could easily add other backends as separate middlewares like Middleware_memory_session and the session API Opium.Session stays the same

In the PR, `Session.{get,set}` is hardcoded to `Middleware_cookie_session.{get,set}`, so think if we want to make it modular, the API would have to break somehow. Not a big deal, but could be worth thinking of it while we're at it. Wouldn't want to block the merge for this though, so if you prefer us to do it as a second step, that's fine by me.

> my plan was to submit a cookie based flash message implementation for Opium and switch to that

Sounds great!

tmattio:
> read/write the cookie from the request/response directly

You'd have to pass the `Cookie.Signer.t` when calling `{get,set}`, which isn't great, so maybe you'd want some kind of functor, which would fit with a modular backend design too.

jerben:
> That's what I meant, sorry. That's why I'm not sure to understand the need for the middleware - couldn't Session.set and Session.get read/write the cookie from the request/response directly?

My plan was to move `Session.t` and `Env.key` to `Opium.Session` eventually and have a dependency on it from the middleware implementations. The API is in `Opium.Session` while the session persistence happens in the middleware.

tmattio:
> The API is in Opium.Session while the session persistence happens in the middleware.

I'm starting to think there's something obvious I'm not seeing, but: why couldn't the persistence happen in `Opium.Session`?

jerben:
> I'm starting to think there's something obvious I'm not seeing, but: why couldn't the persistence happen in Opium.Session?

It could and maybe it should :) I am just using middlewares to swap implementations.. If you give me some more pointers I can try to come up with something else.

tmattio:
> It could and maybe it should :) I am just using middlewares to swap implementations.. If you give me some more pointers I can try to come up with something else.

I'm not against middlewares, but it does come with some drawbacks:

- The users have to remember to add the session middlewares in their routers, otherwise, the sessions can't be used.
- It's harder to think about the code: when you call `Session.set`, what happens? e.g. If you overwrite the cookie with the same key, will it really be overwritten or will the session be stored anyways?

So I tend to prefer keeping things outside of middlewares when possible.

In the case of sessions, I liked the design of https://github.com/inhabitedtype/ocaml-session (although admittedly I found the names of the functions confusing). We can't implement cookie backends with it, but we could take some inspiration from it.

If we want to hide the functors from users, we could imagine an API like this:

```ocaml
let (module User_session) = Opium.Session.signed_cookie ~signer:Config.user_session_signer

(* In the handler *)

User_session.get "_session" request
```

Or, if we accept to require users to have to set their app for sessions to work:
```ocaml
Opium.Session.set_default_signer signer

(* In the handler *)

Opium.Session.Cookie.get "_session" request
```

tmattio:
That being said, most frameworks I have used require a middleware for sessions to work, I'm failing to understand the reasoning behind it.

jerben:
I updated the PR and contributed some tests that shows the usage.

I agree with the limitations of middlewares. This is a simple way to associate 4KB of data to a user session. It is completely up to the user to store a reference such as `user_id` and use some key-value store to persist actual session data.

All you need to understand for Opium are `Request.t -> Response.t Lwt.t` and `Rock.Handler.t -> Rock.Handler.t`, which is quite simple. Shouldn't Opium focus on these abstractions and leave things with lifecycles and (mutable) state to users or higher-level frameworks?

jerben:
We could also read from `Opium.Request.t` and write to `Opium.Response.t`directly without using a middleware, then passing configuration is a bit awkward as there is no central place where you can do that.

rawleyfowler:
Is this still being worked on?

joseferben:
We implemented this [downstream](https://github.com/oxidizing/sihl/blob/master/sihl/src/web_session.ml), it should be straightforward to use that. I am a bit out of the loop, especially regarding the switch back to cohttp. @rgrinberg How would that impact the request/response API?

If there is interest I could update this PR to use Sihl's implementation.

rgrinberg:
> If there is interest I could update this PR to use Sihl's implementation.

That would be useful for sure.

PR #270, opened 2021-05-17
Add App.verbose, App.debug

reynir:
This makes it possible to run with verbose output and debug output when using `App.start`.


tmattio (approved):

tmattio:
Thanks! Could you add an entry in the Changelog?

reynir:
I added an entry, but it's not very elaborate

anuragsoni:
@reynir Thank you!

PR #287, opened 2022-11-15
Add `to_urlencoded_json` to request

rawleyfowler:
This makes it easy to convert URL encoded to your own types. I implemented this in my own application but I think if this is to be `Sinatra` like it should be included. It is nice to have services/functions that take JSON regardless of whether it came via a Post body, or urlencoded.


rawleyfowler:
I should note, I'd also be willing to implement an `of_urlencoded_json` if it's wanted.

